Issue #13, opened 2020-03-26
Remove all use of BaseException


gordonwatts:
Turns out, in python, that BaseException was created because the language needed something as a catch-all. But it was not meant to be used by anyone doing anything but system-level exception plumbing.

Plan:

- Replace every use of `BaseException` with `Exception` or a custom exception class.

Issue #12, opened 2020-03-20
nest_asyncio needs updating


gordonwatts:
The `nest_asyncio` package provided a nice way to avoid having to deal with mixing async event loops and sync calls. This was really only an issue in the Jupyter environment where it would start an event loop for you, but, in many cases, you really wanted to call the synchronous version because it was easy.

## Approach

After investigation - `nest_asyncio` is playing around in the details of the async system, and as it evolves and python goes forward, this is a dangerous thing to do.

Remove the use, and instead start a thread and run as async in that thread, waiting for the thread to complete. This is much more robust - and works in 3.6, 3.7, and 3.8.

Issue #4, opened 2020-03-07
Get this running under 3.8


gordonwatts:
Currently fails one test - but the failure appears to be in the test code, not in the actual code.

Blocker: all the asyncio stuff (see #12) and the way we were throwing exceptions was not compatible with `nest_asyncio` (see #13).

gordonwatts:
It actually works fine in 3.8.0, 3.8.1. In 3.8.2 it fails, and the failure is due to the code/trick i'm using to allow nested running (e.g. so it can run from python and jupyter notebooks). Another case of python not quite getting it right.

See this github bug report: https://github.com/erdewit/nest_asyncio/issues/20

Once that gets fixed, we will have python 3.8.2 (or better) support.

erdewit:
It is fixed in nest_asyncio v1.3.0.

PR #16, opened 2020-03-29
Base Exceptions, python 3.8, asyncio clean up

gordonwatts:
Three issues are addressed in the PR:

- Clean up use of `BaseException` - #13 
- Get all tests working on python 3.6-3.8 - #4 
- Remove the library `nest_asyncio` and use extra threads - #12 

The last one requires some explanation.

### `asyncio` background

`async` and `await` are python's way of working with co-routines. Normally one has two choices: block the thread while waiting for an I/O operation (like on the network), or have some sort of call-back. The `async/await` stuff allows you to write callbacks, without having to write all the extra work.

To make this work a manager must be in place, one for each thread in python's implementation. This is called the event loop. When you call `await` it will hand your thread off to something else that is ready to do work.

For, I hope, obvious reasons, a thread can have only one manager running at a time.

### The problem

The difficulty is how to mix synchronous programming and `async` programming. If the user just wants to run a query, for example, they should not need to deal with all the async stuff. If you decide all the code you will write is going to be synch (e.g. only call `get_data`), then you are fine. Or if you decide you'll only ever call `get_data_async` then you'll also be fine. But what if you mix? When you call `get_data`, it actually calls `get_data_async` under the covers - which means you now need an event loop.

So, one uses the `asyncio.get_event_loop` to get the event loop, and then runs it. This works fine, unless you already have an event loop you don't know about running.

The place that happens is in a Jupyter notebook. It starts an event loop behind your back. So you have to be somehow sensitive to this so you avoid trying to create a new event loop, and further, figure out how to use the already running event loop.

The package `nest_asyncio` was nice - it made the event loop re-entrant. So you could ask it to run a routine even if it was already running. Python 3.8.2, broke that. In fact, looking at what `nest_asyncio` does, it is by design fragile and working around some special stuff that the python libraries do. It as mostly fixed after a bug report, however, not all the way and I was seeing some instabilities in the tests.

### The solution

The solution is to punt. When a synchronous `get_data` request comes in, a new thread is spun up. We then have complete control over the `event_loop` environment. We can safely spin up a new event loop, use it, shut it down, and release the thread. This is the approach taken in the code in this PR.


servicex/servicex.py
```diff
@@ -239,7 +246,23 @@ def get_data(selection_query: str, datasets: Union[str, List[str]],
     Returns:
         df                  Pandas DataFrame that contains the resulting flat data.
     '''
-    nest_asyncio.apply()
     loop = asyncio.get_event_loop()
-    return loop.run_until_complete(get_data_async(selection_query, datasets, servicex_endpoint,
-                                   image=image))
+    if not loop.is_running():
```
BenGalewsky: Can you explain in the PR what is the problem you are addressing with asynch-io and an overview of what changed?

gordonwatts: I've updated the PR's main text. Please check it out and let me know if that makes sense!

BenGalewsky: That is a *great* explanation. PR approval with gold stars!

gordonwatts:
Thanks!!

PR #17, opened 2020-03-29
Add a max_workers parameter to get_data and get_data_async

gordonwatts:
# Motivation and what this fixes

A physicist would like some control over how many resources are used by `servicex` (until auto-scaling is present). This gives the user access to the number of transformers that will run simultaneously.

- This PR adds the parameter `max_workers` to `get_data` and `get_data_async`, which allow the user to control the number of workers used by ServiceX (see issue #17 ). The amount of code changed is actually quite small.

## Reviewing

- @kyungeonchoi - I've put you on review here to make sure you are happy with the interface.
- @BenGalewsky and @mweinberg2718 - I've put you on the review to make sure the text I'm using to describe `max_workers` is ok. Right now, without autoscaling, `max_workers` is really the number of workers, but eventually, it will either be removed, or it will be an upper limit. If it remains, I'm hoping my description of it is useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,8 @@ If your query is badly formed or there is an other problem with the backend, an 
 
 If you'd like to be able to submit multiple queries and have them run on the ServiceX back end in parallel, it may be best to use the `asyncio` interface, which has the identical signature, but is called `get_data_async`.
 
+For documentation of `get_data` and `get_data_async` see the `servicex.py` source file.
+
 # Features
 
 Implemented:
@@ -83,7 +85,6 @@ This code has been tested in several environments:
 
 - Windows, Linux, MacOS
 - Python 3.6, 3.7, 3.8
-   - 3.8.0 and 3.8.1 only. Unfortunately, 3.8.2 has caused `nest_asyncio` to fail. Until that package is updated we are stuck at 3.8.1.
 - Jupyter Notebooks (not automated), regular python command-line invoked source files
 
 # Development
--- a/servicex/servicex.py
+++ b/servicex/servicex.py
@@ -138,10 +138,11 @@ async def _download_new_files(files_queued: Iterable[str], end_point: str,
 async def get_data_async(selection_query: str, datasets: Union[str, List[str]],
                          servicex_endpoint: str = 'http://localhost:5000/servicex',
                          data_type: str = 'pandas',
-                         image: str = 'sslhep/servicex_xaod_cpp_transformer:v0.2') \
+                         image: str = 'sslhep/servicex_xaod_cpp_transformer:v0.2',
+                         max_workers: int = 20) \
         -> Union[pd.DataFrame, Dict[bytes, np.ndarray]]:
     '''
-    Return data from a query with data sets
+    Return data from a query with data sets.
 
     Arguments:
         selection_query     `qastle` string that specifies what columnes to extract, how to format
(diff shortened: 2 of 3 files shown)


gordonwatts:
Ugh.... turns out this includes lots of changes from another PR. Hold off on reviewing this for now.

gordonwatts:
Ok - I missed something here - looks like @BenGalewsky switch this branch to be based off of https://github.com/ssl-hep/ServiceX_frontend/tree/pr_baseexception_async_fixes, rather than develop. Could you switch it back? It should be based off of develop.

kyungeonchoi:
Hi @gordonwatts, I think it is better to assign the number of workers to each request based on the number of files in each did. And the user can set the maximum workers (to each request). For example, if I set maximum workers to 50 and the did contains 5 files, it sets 5 workers. Or even you can hide the number of workers from the user. I guess we need to get the number of files from `didFinder`.

README.md
```diff
@@ -83,7 +85,6 @@ This code has been tested in several environments:
 
 - Windows, Linux, MacOS
 - Python 3.6, 3.7, 3.8
```
mweinberg2718, line 87: Did 3.8.2 get updated to work with nest_asyncio?

servicex/servicex.py
```diff
@@ -138,10 +138,11 @@ def protected_list_objects(client: Minio, request_id: str):
 async def get_data_async(selection_query: str, datasets: Union[str, List[str]],
                          servicex_endpoint: str = 'http://localhost:5000/servicex',
                          data_type: str = 'pandas',
-                         image: str = 'sslhep/servicex_xaod_cpp_transformer:v0.2') \
+                         image: str = 'sslhep/servicex_xaod_cpp_transformer:v0.2',
```
mweinberg2718, line 141: Is there any value in switching this to the v0.4 tag, just to keep it up to date?

gordonwatts: This is a bigger question - I don't know the right way to do this in ServiceX. Let's start a separate discussion outside of this PR about how to address that - as it might require changes in ServiceX itself.

README.md
gordonwatts: Ha! Good catch! So, after looking more carefully at what was going on, I removed `nest_asyncio` totally. In the end, I put in place what I think is a more robust workaround. `nest_asyncio` patches the runtime for python - so it has to be updated all the time. It broke in 3.8.2 because a small error check was added to the python libraries. Now I just run it in a separate thread if there isn't an event loop already running. That seems much more robust.

gordonwatts:
> Hi @gordonwatts, I think it is better to assign the number of workers to each request based on the number of files in each did. And the user can set the maximum workers (to each request). For example, if I set maximum workers to 50 and the did contains 5 files, it sets 5 workers. Or even you can hide the number of workers from the user. I guess we need to get the number of files from `didFinder`.

@kyungeonchoi - Agreed. Unfortunately, there is no way to do that from the front end API. Instead, that is a modification that needs to be made by @BenGalewsky and @mweinberg2718. Until that happens the user is going to have to set this by hand. So the question becomes - what do we do with this until that happens?

servicex/servicex.py
```diff
@@ -230,10 +241,11 @@ def protected_list_objects(client: Minio, request_id: str):
 def get_data(selection_query: str, datasets: Union[str, List[str]],
              servicex_endpoint: str = 'http://localhost:5000/servicex',
              data_type: str = 'pandas',
-             image: str = 'sslhep/servicex_xaod_cpp_transformer:v0.2') \
+             image: str = 'sslhep/servicex_xaod_cpp_transformer:v0.2',
```
mweinberg2718, line 244: Same question here; should we update to the v0.4 tag?

gordonwatts: Yes. I will do that in a separate PR as I move to the final version.

kyungeonchoi:
> > Hi @gordonwatts, I think it is better to assign the number of workers to each request based on the number of files in each did. And the user can set the maximum workers (to each request). For example, if I set maximum workers to 50 and the did contains 5 files, it sets 5 workers. Or even you can hide the number of workers from the user. I guess we need to get the number of files from `didFinder`.
> 
> @kyungeonchoi - Agreed. Unfortunately, there is no way to do that from the front end API. Instead, that is a modification that needs to be made by @BenGalewsky and @mweinberg2718. Until that happens the user is going to have to set this by hand. So the question becomes - what do we do with this until that happens?

You are correct. This is not something from the frontend. I will add a new Github issue describing the desired feature.

gordonwatts:
@BenGalewsky - I think from our conversation you approved this fellow. I'm going to merge.

Issue #27, opened 2020-04-11
Update docs for all the changes


gordonwatts:
Ops! We need documentation!

Issue #22, opened 2020-04-10
Refactor


gordonwatts:
The code is now complex enough you can't see the high-level of what is going on for all the if-statements. Time to move to a nicer structure which is more modular (and then more unit testing will be possible).

- [x] Run a spell checker on everything in this repo!
- [x] Redesign the front end interface. A different method for each return type (along with its async variant). Pass arguments as some coherent, but typed way.
- [x] We are doing two types of caching. Define a unified cache interface that can be layered, and extract out all the caching bits.
- [x] Check the lengths of all functions for "reasonableness" and complexity.
- [x] Unify the test fixtures (so there is only one `good_transform`, for example)
- [x] Examine all type errors and try to minimize them
- [x] Clean out all TODO's.

Many thanks for @BenGalewsky for many of the suggestions from his code reviews for cleaning this code up.

gordonwatts:
Comment by @BenGalewsky: " I think this API would be much easier to work with if you switched from a module to a class. Most of the values would be passed in with the constructor and so the actual method functions will be simpler and have just a few arguments"

Comment by @BenGalewsky : The get_data_async function is too long

Comment by @BenGalewsky : I think this would be a cleaner interface if each data_type were a different function call. That way the return value would be statically defined. Also you'd eliminate a string argument. This is the argument defining the return type for `get_data`.

Spell check the whole repo

This was a bunch of comments that got put in here to help from the task list for this.

PR #46, opened 2020-06-14
Refactoring and Clean up

gordonwatts:
Refactor the code and general clean up

- #22 Over all refactoring, taking into account the redesign of the API around a objects rather than functions.
- #27 Getting what documentation there is up to date.

## Reviewing

These changes are extensive. While new logic was introduced - and I hope is cleaner - it is very hard to look at deltas. Almost nothing was left untouched.


servicex/servicex_remote.py
```diff
@@ -0,0 +1,183 @@
...
+from minio import Minio, ResponseError
+from backoff import on_exception
+import backoff
+
+from .utils import ServiceXException, ServiceXUnknownRequestID
+
+
+# Low level routines for interacting with a ServiceX instance via the WebAPI
+
+async def _get_transform_status(client: aiohttp.ClientSession, endpoint: str,
```
BenGalewsky, line 16: This would be tidier if it were a class. Session and endpoint (and eventually user credentials) will be properties

gordonwatts: Ok - I'll take a look!

PR #74, opened 2020-07-23
Auth with emails instead of usernames

AndrewEckart:
Brings the frontend in line with ssl-hep/ServiceX_App#37, in which the UserModel was changed to use email addresses for authentication instead of usernames.


cnweaver:
I may not have all of the context for this (and I'm probably a bit late bringing this up), but I wanted to note that 1) this will likely become irrelevant if Globus Auth is used in future, as the Globus ID returned from the OAuth process is then the right identifier to use, and 2) emails are a bad choice for identifiers because they aren't stable (they do change over time, such as when users change institutions).

gordonwatts:
Thanks for this, @AndrewEckart !!

Two things:

- Please update the references to servicex/#37 - that is a very old issue.
- Where is this feature in serviceX? Is it in the develop branch yet? The reason behind this is that this is a breaking change - once this is put in to develop it can't be used against a public instance any longer.

AndrewEckart:
> * Please update the references to servicex/#37 - that is a very old issue.
Sorry, wrong repo.

> * Where is this feature in serviceX? Is it in the develop branch yet? The reason behind this is that this is a breaking change - once this is put in to develop it can't be used against a public instance any longer.
Yes, it's in develop, so it's necessary for the Python frontend to work with any deployment made from the current develop branch.

However, ssl-hep/ServiceX_App#49 is well underway. This will replace email/password auth with Globus Auth, so our efforts are perhaps better focused working towards pushing that through instead.

gordonwatts:
It really depends on what is planned for RC2 - if this should go in or not. I'll start a convo in slack.

gordonwatts:
Ok - as far as I can tell, it isn't clear when this will go into a release, so let's have this sit here until we know more about how this is going to work. The changes are pretty simple, so can be pulled in when we want.

gordonwatts:
Ok - time to put this in, finally.

PR #77, opened 2020-07-30
Support environment variables in the config strings

### LINKED ISSUE #73: Add ability to use an ENV variable as the password
As a developer and tester I'd like to use GitHub CI to run test queries against a public instance of `ServiceX`. I'd like the secret information encoded in GitHub secrets.

- [x] Update documentation

gordonwatts:
Main goal: Be able to run tests against a public instance of `ServiceX` in a github CI workflow. Secrets will store the username/password information.

Fixes #73

- Replace all text fetches from the config with `as_str_expanded`, which will expand env var references
- Updated the documentation.


gordonwatts:
This PR is pretty simple - so hopefully it takes you all no time to check it out.

PR #81, opened 2020-08-04
Authenticate with ServiceX API token

AndrewEckart:
Companion to ssl-hep/ServiceX_App#49, which updates the API server to use "ServiceX API tokens", which are JWT refresh tokens issued by us and collected by the user on the web.

This PR updates the `.servicex` config file and `ServiceXAdaptor` to use these long-lived refresh tokens in place of username/email and password. Unit tests have been updated accordingly.


gordonwatts:
This looks good - but is not RC2, right? Where as the email is what is in RC2.

PR #84, opened 2020-08-05
Bring ServiceX Frontend inline with ServiceX 1.0 RC2

### LINKED ISSUE #30: Properly detect a DID not found
The error due to a bogus DID isn't properly bubbled up to the user. Not yet obvious, either, how servicex responds

gordonwatts:
RC2 of ServiceX released a bunch of features. This PR adds features to make them accessible to the user:

- #79 - Fetches the /transformation endpoint data
- #80 - Uses minio auth information from the /transformation endpoint data
- #82 - When a file fails to transform, dump transform logs of failed pods to the logger.warning.
- #30 - If the DID look up fails (or simlar error), report error message to use from /transform status
- ServiceX_App/34 - The ServiceX app now uses email/password to authenticate.

Goal would be to merge this along with other changes into master as version 1.1 once approved.

### DIFF
--- a/servicex/cache.py
+++ b/servicex/cache.py
@@ -2,7 +2,7 @@ import json
 from pathlib import Path
 from typing import Dict, List, Optional, Tuple, Any
 
-from .utils import _query_cache_hash, sanitize_filename
+from .utils import ServiceXException, _query_cache_hash, sanitize_filename
 
 
 class Cache:
@@ -81,6 +81,8 @@ class Cache:
 
         '''
         f = self._query_status_cache_file(request_id)
+        if not f.exists():
+            raise ServiceXException(f'Not cache information for query {request_id}')
         with f.open('r') as o:
             return json.load(o)
 
--- a/servicex/cache.py
+++ b/servicex/cache.py
@@ -38,6 +38,10 @@ class Cache:
         h = _query_cache_hash(json)
         return self._path / 'query_cache' / h
 
+    def _query_status_cache_file(self, request_id: str) -> Path:
+        'Return the query cache file'
+        return self._path / 'query_cache_status' / request_id
+
     def _files_cache_file(self, id: str) -> Path:
         'Return the file that contains the list of files'
         return self._path / 'file_list_cache' / id
@@ -55,6 +59,30 @@ class Cache:
         f.parent.mkdir(parents=True, exist_ok=True)
         with f.open('w') as o:
             o.write(f'{v}\n')
+
+    def set_query_status(self, query_info: Dict[str, str]):
+        '''Cache a query status (json dict)
+
+        Args:
+            query_info (Dict[str, str]): The info we should cache. Must contain `request_id`.
+        '''
+        assert 'request_id' in query_info, \
+            "Internal error - request_id should always be part of info returned"
+        f = self._query_status_cache_file(query_info['request_id'])
+        f.parent.mkdir(parents=True, exist_ok=True)
+        with f.open('w') as o:
+            json.dump(query_info, o)
+
+    def lookup_query_status(self, request_id: str) -> Dict[str, str]:
+        '''Returns the info from the last time the query status was cached.
+
+        Args:
+            request_id (str): Request id we should look up.
+
+        '''
+        f = self._query_status_cache_file(request_id)
+        with f.open('r') as o:
+            return json.load(o)
 
     def remove_query(self, json: Dict[str, str]):
         f = self._query_cache_file(json)
--- a/servicex/minio_adaptor.py
+++ b/servicex/minio_adaptor.py
@@ -147,7 +147,7 @@ class MinioAdaptorFactory:
         if self._config_adaptor is not None:
             logging.getLogger(__name__).debug('Using the config-file minio_adaptor')
             return self._config_adaptor
-        raise Exception("Do not know how to create a Minio Login info")
+        raise ServiceXException("Do not know how to create a Minio Login info")
 
     def _from_config(self, c: ConfigView) -> MinioAdaptor:
         '''Extract the Minio config information from the config file(s). This will be used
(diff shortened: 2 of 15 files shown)


gordonwatts:
Sorry - this is a big PR - has a bunch of new features and code changes in it!

servicex/cache.py
```diff
@@ -56,6 +60,30 @@ def set_query(self, json: Dict[str, str], v: str):
...
             o.write(f'{v}\n')
 
+    def set_query_status(self, query_info: Dict[str, str]):
+        '''Cache a query status (json dict)
+
+        Args:
+            query_info (Dict[str, str]): The info we should cache. Must contain `request_id`.
+        '''
+        assert 'request_id' in query_info, \
+            "Internal error - request_id should always be part of info returned"
```
AndrewEckart, line 70: Assertions [should not be used for error handling](https://stackoverflow.com/questions/20076182/what-are-the-implications-of-using-assert-vs-raise-exception). Instead, we should raise a custom exception.

tests/test_minio_adaptor.py
```diff
@@ -193,13 +206,41 @@ def test_minio_adaptor_factory_use_api_usernamepassword():
...
 
-    m = minio_adaptor_factory(c)
+    m = MinioAdaptorFactory(c).from_best()
     assert m._endpoint == 'the-good-host.org:9000'
     assert m._access_key == "thegoodplace"
     assert m._secretkey == "forkingshirtballs!"
 
 
-def test_minio_adaptor_factory_use_default_username_password():
+def test_factory_use_default_username_password():
```
AndrewEckart, line 215: Should be named `test_factory_use_default_email_password`, my bad.

servicex/minio_adaptor.py
```diff
@@ -116,6 +100,84 @@ def do_copy() -> None:
...
+                    and 'minio-access-key' in transation_info \
+                    and 'minio-secret-key' in transation_info:
+                logging.getLogger(__name__).debug('Using the request-specific minio_adaptor')
+                return MinioAdaptor(transation_info['minio-endpoint'],
+                                    transation_info['minio-access-key'],
+                                    transation_info['minio-secret-key'])
+        if self._config_adaptor is not None:
+            logging.getLogger(__name__).debug('Using the config-file minio_adaptor')
+            return self._config_adaptor
+        raise Exception("Do not know how to create a Minio Login info")
```
AndrewEckart, line 150: Best practice is to define a custom exception class which subclasses `Exception`.

gordonwatts: Yeah - probably right - this is me being lazy. I'll update!

servicex/cache.py
gordonwatts: I've been using them like this:

- These are things that should never happen, but if they do, they will be hard to figure out why they happened from the user end.
- They require a whole extra test to provide code-coverage if I use an if statement.

I could just remove it: after all, we never expect this.

AndrewEckart: Okay, that's a perfectly good way to use them (I wasn't sure how rare this error was). It's what the article I linked describes as a sanity check. For example, if you had a function `square = lambda x: x*x` then you might do:
```
y = square(x)
assert y >= 0
```
which should never execute if x is a real number.

BenGalewsky:
Would prefer multiple PRs, but here goes...

servicex/servicex.py
```diff
@@ -37,7 +37,7 @@ class ServiceXDataset(ServiceXABC):
     '''
     def __init__(self,
                  dataset: str,
-                 image: str = 'sslhep/servicex_func_adl_xaod_transformer:130_reset_cwd',
+                 image: str = 'sslhep/servicex_func_adl_xaod_transformer:29_dont_truncate_error_logs',  # NOQA
```
BenGalewsky, line 40: Should this be 1.0.0-RC.2 for now?

AndrewEckart:
Comment on versioning - does this need to be version 1.1? Can we align the version numbers between the backend and the frontend?

gordonwatts:
> Looks good apart from the hardcoded docker image tag in servicex.py

If ypu know how that should be handled, i would be up for that!

gordonwatts:
> Comment on versioning - does this need to be version 1.1? Can we align the version numbers between the backend and the frontend?

Ideally. We will try to get there.

gordonwatts:
> Would prefer multiple PRs, but here goes...

Yeah
I am up for that in the future, but that means more for you to review!

AndrewEckart:
> > Would prefer multiple PRs, but here goes...
> 
> Yeah
> I am up for that in the future, but that means more for you to review!

This shouldn't be a problem with our new branching strategy. Each PR should be merged directly into develop, and reviewed by anyone necessary at that time. I don't think we'll need any more aggregate PRs like this.

servicex/minio_adaptor.py
gordonwatts: Ok - PR to remove all Exception throws in the main code! Thanks for the suggestion!

servicex/servicex.py
gordonwatts: This will get fixed as soon as there is a working tag of that name! So not in time for this PR, but shortly after that, I hope!

gordonwatts:
> This shouldn't be a problem with our new branching strategy. Each PR should be merged 
> directly into master (or whatever we call the trunk branch), and reviewed by anyone 
> necessary at that time. I don't think we'll need any more aggregate PRs like this. 

This won't change the fundamental - the issue is that if there are too many PR's then it takes too much of other people's time.

gordonwatts:
> This will get fixed as soon as there is a working tag of that name! So not in time for this PR, but shortly after that, I hope! 

The new version was released, so this has been updated to the `v1.0.0-rc.2` tag.

PR #103, opened 2020-09-18
Allow multiple endpoints in a .servicex file

### LINKED ISSUE #101: Add "types" to the front-end to allow for different servicex backends to be accessed
As a user of `servicex`, I want to be able to use `uproot` and `xAOD` queries without having to make changes to my file system `.servicex` file or my code.

## Assumptions

1. User is using the `.servicex` file to provide credential and end-point information
1. User is trying to run a query on the `xAOD` backend
1. User is trying to run a query on the `uproot` backend.

## Acceptance

The user should not need to alter their `.servicex` file between runs of the two backends - they should be able to run them simultaneously, even.

## The Problem

Currently, there are different backends for `servicex` depending on the type of file you want to go after - be they `uproot`, `xAOD`, or CMS `treemaker`, etc. The easiest way to store credentials is currently in the `.servicex` file:

The current `servicex` file looks like the following:

```
api_endpoint:
  endpoint: http://xaod-servicex.uc.ssl-hep.org/
  email: gwatts@uw.edu
  password: forkingloveit!-fake-
```

This file has to be updated in order to run on more than one backend.

## Proposal

Alter the `servicex` file to look like the following (for `uproot` and `xaod`):

```
api_endpoins:
  - api_endpoint:
    type: xaod
    endpoint: http://xaod-servicex.uc.ssl-hep.org/
    email: gwatts@uw.edu
    password: forkingloveit!-fake-
  - api_endpoint:
    type: uproot
    endpoint: http://uproot-servicex.uc.ssl-hep.org/
    email: gwatts@uw.edu
    password: forkingloveit!-fake-
```

Further, modify the creation API of the `ServiceXDataset` to look like the following:

```
    def __init__(self,
                 dataset: str,
                 backend_type: Optional[str] = None,
                 image: str = 'sslhep/servicex_func_adl_xaod_transformer:v1.0.0-rc.2',  # NOQA
                 max_workers: int = 20,
                 servicex_adaptor: ServiceXAdaptor = None,
                 minio_adaptor: Union[MinioAdaptor, MinioAdaptorFactory] = None,
                 cache_adaptor: Optional[Cache] = None,
                 status_callback_factory: Optional[StatusUpdateFactory] = _run_default_wrapper,
                 local_log: log_adaptor = None,
                 session_generator: Callable[[], Awaitable[aiohttp.ClientSession]] = None,
                 config_adaptor: ConfigView = None):
```

The change is the addition of `backend_type`, and it is required if `servicex_adaptor` is null. It can be any string, and it has to match with some entry in the `.servicex` file unless the `servicex_adaptor` has been passed in - which will override everything since the user has taken responsibility for creating it.

Usage: in the case of the `func_adl_servicex` it contains an object called `ServiceXDatasetSource` and that object knows if it is aiming at a `uproot` or a `and` backend. So when it created the `ServiceXDataset` it would know the type to send at this `ServiceXDataset` object.

## Future

The logic for reading in the `.servicex` and overriding parameter in the `ServiceXDataset` constructor is starting to get complex and is being spread around in several files. Time to refactor and create a low level `ServiceXDataset` object that has no optional arguments. And then a higher-level one whose sole job is to read in defaults and create the low-level one. This will keep the logic in one easy-to-see place, and simplify testing (though some test harnesses will get more complex).

## Progress

- [ ] Add tests
- [ ] Update any documentation needed

gordonwatts:
- Update the .service file parsing (as per #101).
- Add extra error messages to help user understand what has gone wrong if mis-spefied.
- Update documentation on what the `.servicex` file should look like.
- Update scripts on working servicex stuff.

One remaining item is to double check this works by running it against a xaod backend. Will do that as soon as I've got the name of the thing.

Fixes #101

This is a breaking change! The `.servicex` file has changed. And the ctor for the `ServiceXDataset` has changed!!


gordonwatts:
This is ready for a quick review. Make sure to read the issue #101 to get an idea of what I'm trying to do here. Even that approach is up for discussion here.

BenGalewsky:
Seems like a good idea... is it too complicated to say that if you only need to specify the backend if there are more than one specified in .servicex?

BenGalewsky (approved):

d-w-liu (approved):

gordonwatts:
> Seems like a good idea... is it too complicated to say that if you only need to specify the backend if there are more than one specified in .servicex?

If that is the case and then you try to use another backend without updating the, you'll get some pretty awful errors - that don't really make sense (we know this because both @d-w-liu and I have struggled through them - we know them now, but that is only because we've been burned).

PR #119, opened 2020-09-26
Add ability to ignore the local cache

### LINKED ISSUE #44: Add `ignore_cache` option so that local cache can be ignored
https://github.com/ssl-hep/ServiceX_frontend/blob/7ba0e81f07efa7ede2465bb867ee5768ba324547/servicex/servicexabc.py#L173-L173

As a user there are times I'd like to force the backend/servicex to re-run even though the data is already locally stored. The plan is to use a with ignore_cache() to implement this.

gordonwatts:
This code adds a context manager that, while in effect, will prevent queires from being found in the cache when they are initiated.

Fixes #44

### DIFF
--- a/servicex/cache.py
+++ b/servicex/cache.py
@@ -1,14 +1,53 @@
 import json
 from pathlib import Path
 from typing import Dict, List, Optional, Tuple, Any
+from contextlib import contextmanager
 
 from .utils import ServiceXException, _query_cache_hash, sanitize_filename
+
+_ignore_cache = False
+
+
+@contextmanager
+def ignore_cache():
+    '''This will cause all caches to be ignored while it is invoked:
+
+    ```
+    with ignore_cache():
+        ServiceXDataset(...).get_data...()
+    ```
+
+    If you want to do this globally, you can just use the `__enter__()` method.
+    This is probably the only way to do this accross cells in a notebook.
+
+    ```
+    i = ignore_cache()
+    i.__enter__()
+    ... Query code, jupyter notebook cells, etc. go here
+    i.__exit(None, None, None)
+    ```
+
+    Note:
+
+    - The only time the cache is checked is when the query is actually made, not when
+      the servicex dataset object is created!
+    - Calls to this can be safely nested.
+    - Note that calling this doesn't clear the cache or delete anything. It
+      just prevents the cache lookup from working while it is in effect.
+    '''
+    global _ignore_cache
+    old_value = _ignore_cache
+    _ignore_cache = True
+    yield
+    _ignore_cache = old_value
 
 
 class Cache:
     '''
     Caching for all data returns from the system. It provides both in-memory
     and on-disk cache.
+
+    TODO: Rename this to be an adaptor, unifying how we name things
     '''
     _in_memory_cache = {}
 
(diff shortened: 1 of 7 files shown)


servicex/cache.py
```diff
@@ -1,14 +1,53 @@
...
+    i.__enter__()
+    ... Query code, jupyter notebook cells, etc. go here
+    i.__exit(None, None, None)
+    ```
+
+    Note:
+
+    - The only time the cache is checked is when the query is actually made, not when
+      the servicex dataset object is created!
+    - Calls to this can be safely nested.
```
AndrewEckart, line 34: It's a nice touch, but what's the use case for nested calls to `ignore_cache`?

AndrewEckart (commented):
The context manager implementation is well-done and pretty flexible, but I think it'd be nice if there was an option (maybe just syntactical sugar) to do it on a per-request basis as a keyword argument, e.g. `ds.value(ignore_cache=True)`. 

The advantage is that if you wanted to make several calls in some specific order with a mixture of caching and no caching for whatever reason, you wouldn't have to write as much boilerplate to jump in and out of the context manager. If that's too much of a hassle, don't worry about it.

In general I think one major use case for ignoring the cache is that as a ServiceX developer you might have one endpoint where you want to skip the cache 100% of the time (say it's a continuously-delivered build of ServiceX based on the develop branch), and another endpoint where you don't want to skip the cache (maybe a production build of ServiceX). So it might also be nice to set it at the level of an entire endpoint, or as an attribute on a dataset's `Cache` adaptor.

servicex/cache.py
gordonwatts: In my last implementation, with LINQ, I had this, and at one point I was deep in the code and wanted to make sure to ignore the cache - for testing - but then left it in for some reason, then did the ignore at a upper level, and I lost about 30 minutes of forking around to figure out what had happened. So now, in general, when I do somethign like this, I always make it nest proof. With the tools python has, this makes it totally trivial, actually.

gordonwatts:
> The context manager implementation is well-done and pretty flexible, but I think it'd be nice if there was an option (maybe just syntactical sugar) to do it on a per-request basis as a keyword argument, e.g. `ds.value(ignore_cache=True)`.
> 
> The advantage is that if you wanted to make several calls in some specific order with a mixture of caching and no caching for whatever reason, you wouldn't have to write as much boilerplate to jump in and out of the context manager. If that's too much of a hassle, don't worry about it.
> 
> In general I think one major use case for ignoring the cache is that as a ServiceX developer you might have one endpoint where you want to skip the cache 100% of the time (say it's a continuously-delivered build of ServiceX based on the develop branch), and another endpoint where you don't want to skip the cache (maybe a production build of ServiceX). So it might also be nice to set it at the level of an entire endpoint, or as an attribute on a dataset's `Cache` adaptor.

Excellent points. What about doing something like:

```
ds = ServiceXDataset(...)
with ds.ignore_cache():
  ds.Select(...)....value()
```

The reason not to put the ignore cache in the value is that that value is often burried in someone else's library. Where as the physicist is always going to have to have some level of direct hand in creating the ServiceXDataset.

You could, I suppose, also create the ServiceXDataset with an option to always ignore the cache... Ok - I like both of these. Going to implement, thank you!!

gordonwatts:
Ok @AndrewEckart - take a look again and let me know what you think!

AndrewEckart (approved):
LGTM!

Issue #128, opened 2020-10-29
Long result filenames cause "Filename too long" Errors


BenGalewsky:
CMS nanoAOD filenames can be quite long. When they are downloaded from Minio, they throw an exception:
```
OSError: [Errno 63] File name too long: 'cache/servicex/data/5a73ce2d-0f4a-4f33-b81b-b40454c9eff4/root___xrootd-local.unl.edu_1094__store_mc_RunIIAutumn18NanoAODv7_DYJetsToLL_M-50_TuneCP5_13TeV-amcatnloFXFX-pythia8_NANOAODSIM_Nano02Apr2020_102X_upgrade2018_realistic_v21_ext2-v1_100000_03965A2E-4DA7-5444-B325-F18BDA18787C.root.parquet.temp3246c4bac2284c1669525077fa22685d.part.minio'
```

PR #129, opened 2020-10-29
Truncate filenames to avoid ones that are too long

BenGalewsky:
# Problem
CMS NanoAOD files can wind up with really long names. This causes errors when they are downloaded from Minio.

This is solution to #128 

# Approach
* Establish a `MAX_PATH_LEN`
* In the Minio adaptor truncate the filename from the left to keep (when combined with the parent path) within the limit.

This assumes that for a dataset, the rightmost characters are most unique.

### DIFF
--- a/servicex/cache.py
+++ b/servicex/cache.py
@@ -1,6 +1,4 @@
 import json
-from os import path
-
 from pathlib import Path
 from typing import Dict, List, Optional, Tuple, Any
 from contextlib import contextmanager
@@ -184,10 +182,7 @@ class Cache:
         OSError: [Errno 63] File name too long error Assume that the most unique part of
         the name is the right hand side
         '''
-        parent = Path(path.join(self._path, 'data', request_id))
+        parent = self._path / 'data' / request_id
         parent.mkdir(parents=True, exist_ok=True)
         sanitized = sanitize_filename(data_name)
-        return Path(path.join(
-            parent,
-            sanitized[-1 * (MAX_PATH_LEN - len(parent.name)):]
-        ))
+        return parent / sanitized[-1 * (MAX_PATH_LEN - len(parent.name)):]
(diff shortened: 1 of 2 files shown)


servicex/cache.py
```diff
@@ -175,6 +180,14 @@ def data_file_location(self, request_id: str, data_name: str) -> Path:
...
         '''
-        (self._path / 'data' / request_id).mkdir(parents=True, exist_ok=True)
-        return self._path / 'data' / request_id / sanitize_filename(data_name)
+        parent = Path(path.join(self._path, 'data', request_id))
+        parent.mkdir(parents=True, exist_ok=True)
+        sanitized = sanitize_filename(data_name)
+        return Path(path.join(
+            parent,
+            sanitized[-1 * (MAX_PATH_LEN - len(parent.name)):]
+        ))
```
gordonwatts, line 193: Any reason not to use 

```
return parent / sanitized[...]
```
and not include the path module here?

servicex/cache.py
```diff
@@ -175,6 +180,14 @@ def data_file_location(self, request_id: str, data_name: str) -> Path:
         '''
         Return the path to the file that should be written out for this
         data_name. This is where the output file should get stored.
+        Truncate the leftmost characters from filenames to avoid throwing a
+        OSError: [Errno 63] File name too long error Assume that the most unique part of
+        the name is the right hand side
         '''
-        (self._path / 'data' / request_id).mkdir(parents=True, exist_ok=True)
-        return self._path / 'data' / request_id / sanitize_filename(data_name)
+        parent = Path(path.join(self._path, 'data', request_id))
```
gordonwatts, line 187: Same comment here. Use the line from 179

gordonwatts: (same comment as below - use the Path object?

gordonwatts (commented):
Update the approach above to reflect that we are change the cache.

Are there any tests that tickle the "too long" pathname condition?

PR #136, opened 2020-12-06
Streaming Data Interface

gordonwatts:
This PR adds streaming a set of methods to `servicex` that return streaming data:

- Use the `async` generator pattern to return python data objects
- Each data object contains a servicex filename and the type of data requested (awkward, pandas, root-file path, pandas file path, or minio-url)
- The minio-url is new: that returns a URL that the user can use to access the data directly in the minio S3 store hosted by ServiceX.


gordonwatts:
@BenGalewsky - I checked it still "works", and added some doc info to the readme file. This contains all the streaming stuff - I need to get this cleared out of the way before tackling other things because of the way git works ("clearing the decks so-to-speak"). I'll give you till Monday afternoon to ask for a hold or approve, or I'm going to merge it with master so I can get onto the smaller stuff that is holding up others. :-) :-)

BenGalewsky (approved):
Do it!

PR #137, opened 2020-12-06
Allow pyarrow v1 or 2

gordonwatts:
Allow `pyarrow` 2.0 to be installed


BenGalewsky (approved):
Just like it says on the tin!

PR #138, opened 2020-12-06
Get config information from servicex.yaml and servicex.yml and .servicex files

gordonwatts:
With this PR, the frontend will search for config information in a `servicex.yaml`, `servicex.yml`, and `.servicex` file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,11 +138,11 @@ As mentioned above, the `.servicex` file is read to pull a configuration. The se
 1. Your current working directory
 2. Your home directory
 
-The file can be named any of the following:
+The file can be named any of the following (ordered by precedence):
 
-- `.servicex`
 - `servicex.yaml`
 - `servicex.yml`
+- `.servicex`
 
 The file can contain an `api_endpoint` as mentioned above. In addition the other following things can be put in:
 
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Before you can use this library you'll need:
 
 The API access information is normally placed in a `.servicex` file (to keep this confidential information form accidentally getting checked into a public repository). The `servicex` library searches for configuration information in several locations to determine what end-point it should connect to, in the following order:
 
-1. A `.servicex` file in the current working directory
+1. A `.servicex` file in the current working directory (it can also be named `servicex.yaml` or `servicex.yml`)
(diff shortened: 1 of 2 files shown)


AndrewEckart (approved):
What happens when more than one of `.servicex`, `servicex.yaml`, and `servicex.yml` are present? I tested and it seems that `servicex.yaml` takes precedence over `.servicex`, at least in the local directory. Maybe this is related to the ordering of the calls to `_add_from_path`? If so, the ordering should be consist and the docs should reflect it.

If supporting multiple filenames is undesirably complex, my suggestion would be that we deprecate `.servicex` and support only `servicex.yaml` moving forward (I believe the `.yml` extension has been discouraged for a while). While we'd lose the "hidden" behavior of the dot prefix, this would be a bit simpler; YAML files are widely used for configuration, and our current userbase is small enough for the impact of a breaking change to be minimal.

That said, I'm happy with keeping 2 or 3 reserved filenames, as long as it's clear which one will be used first. The YAML parsing seems to be working great.

gordonwatts:
Thanks for looking. I'll update the docs to reflect the ordering explicit. It was done that way on purpose, actually - Make sure the "visible" file has precedence over the "invisible" file. That makes debugging a little easier - one less thing to think about.

I'll keep both yml and yaml - I see both around.

I'll keep the hidden file. One's Linux home directory is already cluttered enough without having to liter it with config files. It seems like it is the standard for linux tools anyway.

AndrewEckart:
> Thanks for looking. I'll update the docs to reflect the ordering explicit. It was done that way on purpose, actually - Make sure the "visible" file has precedence over the "invisible" file. That makes debugging a little easier - one less thing to think about.
> 
> I'll keep both yml and yaml - I see both around.
> 
> I'll keep the hidden file. One's Linux home directory is already cluttered enough without having to liter it with config files. It seems like it is the standard for linux tools anyway.

Great - my suggestion is that the ordering between `.yaml` and `.yml` on lines 24-25 and 37-38 should be the same (probably `.yaml` before `.yml`). Thanks for the quick turnaround!

gordonwatts:
Ok! I'll later that. As far as quick turn around... given when you submitted this bugreport, I'd not call this quick! :-)

PR #145, opened 2021-01-21
move from .servicex to servicex.yaml

ivukotic:
just a documentation change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,16 +26,16 @@ Before you can use this library you'll need:
 
 ### How to access your endpoint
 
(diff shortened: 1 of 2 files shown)


gordonwatts:
As a quick note - both filenames are valid. I don't think we have to advertise them both - but they both work. the .servicex is nice because it disappears in your directory and allows you to keep it uncluttered.

gordonwatts:
Closing this in favor of https://github.com/ssl-hep/ServiceX_frontend/pull/156

PR #148, opened 2021-02-19
Improve error message when Minio is unreachable

AndrewEckart:
Currently, if a request is submitted and Minio cannot be reached, it fails with the following error message:
```
    raise MaxRetryError(_pool, url, error or ResponseError(cause))
urllib3.exceptions.MaxRetryError: HTTPConnectionPool(host='localhost', port=9000): Max retries exceeded with url: /0a5d544d-7dd9-4ea8-87e5-eda66a100684?location= (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fabef32eb20>: Failed to establish a new connection: [Errno 111] Connection refused'))
```

This networking exception is not particularly informative. It's clear that we failed to make a connection to `localhost:9000/0a5d544d-...`, but it's not clear what's located at that URL. Combing through the stack trace reveals that it's a Minio URL.

With this PR, the error message would be:
```
    raise ServiceXException(msg)
servicex.utils.ServiceXException: (ServiceXException(...), 'Unable to reach Minio at localhost:9000/44d2056d-2ca1-46c2-8825-14b12b42e6e6?location=. Max retries exceeded.')
```
This makes it more clear that the connection failure occurred when trying to reach Minio.


gordonwatts:
Nice! This is better!

gordonwatts:
Added a minor test for coverage. I think this might be important because I suspect this is one thing we'll have to look at in the future when we start to see intermitent failures. For example, a glitch because you are on wifi and move from one coffee shop to the next. If you have a "long-running" calculation, it would be bad to tear the whole thing down.

PR #149, opened 2021-03-15
Update qastle 0.10

BenGalewsky:
# Problem
Qastle library has been upgraded to 0.10 we need to update the frontend to pull this in otherwise we can get incompatible requirements.

Also, a change in the `idna` library breaks the `requests` dependency. As shown in https://github.com/psf/requests/issues/5710

# Approach
It seems that for a library of libraries like this one it makes sense to be generous with dependency requirements and not pin them down too tightly. Changed the requirement to be `qastle>=0.10`


setup.py
```diff
@@ -37,13 +37,14 @@
       python_requires='>=3.6',
       test_suite="tests",
       install_requires=[
+          "idna==2.10",  # Required to thread version needle with requests library
```
gordonwatts, line 40: Is this going to come back and bite us in the butt at some later time?

BenGalewsky: Added safety check to make sure the bite isn't fatal

PR #156, opened 2021-03-29
Fix up references to .servicex files in documentation and comments

### LINKED ISSUE #155: Make references to servicex.yaml more explicit in the documentation
This is to track the issues that were @ivukotic was trying to fix or improve in his pull request #145.

- Keep references to .servicex
- Make sure they are paired with servicex.yaml as well.

gordonwatts:
We now use `servicex.yaml`, `servicex.yml`, and `.servicex` files for configuration. The documentation and code comments hadn't quite caught up with that. This should take care of that.

Fixes #155

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,16 +26,17 @@ Before you can use this library you'll need:
 
 ### How to access your endpoint
 
-The API access information is normally placed in a `.servicex` file (to keep this confidential information form accidentally getting checked into a public repository). The `servicex` library searches for configuration information in several locations to determine what end-point it should connect to, in the following order:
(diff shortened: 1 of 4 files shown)


ivukotic (commented):
in Readme.md line 29, there is "file file".
Would have been nice to explain "caching" in a bit more details, people could get surprised to find their jupyterlab running in k8s being terminated due to "disk pressure" since their pod filled up ephemeral storage, since they accessed more then eg. 5GB of data from servicex.
Otherwise it is fine.

gordonwatts:
Ok - that is actually quite interesting.

The current cache is quite simplistic - it just fills up. Perhaps a better layer is needed. Or perhaps we should have an option to turn it off if you are very close to servicex.

PR #157, opened 2021-03-30
Default data cache location is per-user

### LINKED ISSUE #147: Make default location of the cache unique for a user
We had multiple users try the frontend on our group's server, and immediately had an issue in that the cache directory under /tmp was by default the same for all, but permissions were only set up to permit one of them to use it properly. It is possible of course to override the cache location, but it seems like it would be a UX improvement to make the cache directory user-specific (of course there might be people who want to have a shared cache but that seems like an advanced option).

gordonwatts:
- The default cache location on linux is now `/tmp/servicex_<user>`, and on windows similar, but in the `TEMP` directory (which is already per-user).
- Also updated the docs to note that on linux the `/etc/servicex` directory is searched for a file since the `confuse` python config library is getting used.

Fixes #147

### DIFF
--- a/servicex/config_default.yaml
+++ b/servicex/config_default.yaml
@@ -5,8 +5,9 @@ api_endpoints:
   - endpoint: http://localhost:5000
     # token: xxx
 
-# This is the path of the cache. The "/tmp" will be translated, platform appropriate.
-cache_path: /tmp/servicex
+# This is the path of the cache. The "/tmp" will be translated, platform appropriate, and
+# the env variable USER will be replaced.
+cache_path: /tmp/servicex_${USER}
 
 # This is a dummy value, here only to make sure that unit testing
 # works properly before package release.
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ For documentation of `get_data` and `get_data_async` see the `servicex.py` sourc
 
 ### The Local Cache
 
-To speed things up - especially when you run the same query multiple times, the `servicex` package will cache queries data that comes back from Servicex. You can control where this is stored with the `cache_path` in the `.servicex` file (see below).
+To speed things up - especially when you run the same query multiple times, the `servicex` package will cache queries data that comes back from Servicex. You can control where this is stored with the `cache_path` in the configuration file (see below).
 
 There are times when you want the system to ignore the cache when it is running. You can do this by using `ignore_cache()`:
 
@@ -138,6 +138,7 @@ As mentioned above, the `.servicex` file is read to pull a configuration. The se
 1. Your current working directory
 2. Any working directory above your current working directory.
 3. Your home directory
+4. Any other directory searched by the python [`confuse`](https://confuse.readthedocs.io/en/latest/usage.html#search-paths) library. This includes a machine's `/etc` directory on Linux.
 
 The file can be named any of the following (ordered by precedence):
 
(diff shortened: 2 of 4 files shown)


gordonwatts:
@ponyisi - I can't tag you in this for some reason as a reviewer, please do have a look however, and leave comments!

servicex/config_default.yaml
```diff
@@ -5,8 +5,9 @@ api_endpoints:
   - endpoint: http://localhost:5000
     # token: xxx
 
-# This is the path of the cache. The "/tmp" will be translated, platform appropriate.
-cache_path: /tmp/servicex
+# This is the path of the cache. The "/tmp" will be translated, platform appropriate, and
+# the env variable USER will be replaced.
+cache_path: /tmp/servicex_${USER}
```
AndrewEckart, line 10: Another option is `.../servicex/${USER}`, to avoid cluttering the parent directory if ServiceX is being used by a large number of users on the same machine.

AndrewEckart (commented):
I'm not sure about the choice of `/tmp` here. The [FHS](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) includes the following directories for temporary files:

/tmp | Directory for temporary files (see also /var/tmp). Often not preserved between system reboots and may be severely size-restricted.
-- | --
/var/tmp | Temporary files to be preserved between reboots

It seems like `/var/tmp` might be a better fit.

servicex/config_default.yaml
gordonwatts: I thought of that, but that means whoever creates the `/tmp/servicex` directory has to get the permissions right - so we are back with the same issue that started this.

gordonwatts:
> I'm not sure about the choice of `/tmp` here. The [FHS](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) includes the following directories for temporary files:
> 
> /tmp	Directory for temporary files (see also /var/tmp). Often not preserved between system reboots and may be severely size-restricted.
> /var/tmp	Temporary files to be preserved between reboots
> It seems like `/var/tmp` might be a better fit.

I'd never heard of /var/tmp. As far as I can tell, this isn't a POSIX standard, just redhat, so I'm not so sure moving to it is the right thing to do. Also, /var/tmp gets cleaned in many systems too.

So, I'll leave it like this as the default for now. A user can switch it, and if there is a real need for this - we can look into doing something a lot more complex, like a list of possible tmp directories. But that is adding complexity I'd like to avoid. Or if another need comes up...

AndrewEckart:
> > I'm not sure about the choice of `/tmp` here. The [FHS](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) includes the following directories for temporary files:
> > /tmp	Directory for temporary files (see also /var/tmp). Often not preserved between system reboots and may be severely size-restricted.
> > /var/tmp	Temporary files to be preserved between reboots
> > It seems like `/var/tmp` might be a better fit.
> 
> I'd never heard of /var/tmp. As far as I can tell, this isn't a POSIX standard, just redhat, so I'm not so sure moving to it is the right thing to do. Also, /var/tmp gets cleaned in many systems too.
> 
> So, I'll leave it like this as the default for now. A user can switch it, and if there is a real need for this - we can look into doing something a lot more complex, like a list of possible tmp directories. But that is adding complexity I'd like to avoid. Or if another need comes up...

Here is a post on Unix/Linux StackExchange with some more context: https://unix.stackexchange.com/questions/30489/what-is-the-difference-between-tmp-and-var-tmp

You are correct that `/var/tmp` is not in the POSIX standard, but I wouldn't characterize it as just a Red Hat thing - I believe that every major distribution is FHS compliant, although I can't find a great source on which aren't (Debian/Ubuntu certainly are).

I think there is a good argument that the utility of the cache is limited if it's not preserved across reboots. But like you said, it's already configurable, so concerned users can change it if needed.

PR #162, opened 2021-04-03
Add URL property to StreamInfoPath

### LINKED ISSUE #161: Give StreamInfoPath a URL property to match the interface of StreamInfoUrl
# Problem
If we switch from `stream_result_file_urls` to `stream_result_files` in the Local executor, we want the interface to the returned objects to be the same

BenGalewsky:
# Problem 
If we switch from `stream_result_file_urls` to `stream_result_files` in the Local executor, we want the interface to the returned objects to be the same

Fixes #161 

# Approach
Added URL property to StreamInfoPath that returns a `file://` URI to the downloaded file

### DIFF
--- a/servicex/servicex.py
+++ b/servicex/servicex.py
@@ -100,6 +100,15 @@ class StreamInfoPath(StreamInfoBase):
         '''
         return self._path
 
+    @property
+    def url(self) -> str:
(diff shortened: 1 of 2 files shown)


gordonwatts:
Looks great - I'll merge it now!

PR #205, opened 2021-09-10
Allow passing a title for `get_data_X` functions

kyungeonchoi:
Due to the condition on the length of arg, it crashes when passing a title to the `get_data_X` functions. The fix allows an additional argument for a title.


kyungeonchoi:
@gordonwatts - could you have a look this? I want to avoid below screenshot :) 
Screen Shot 2021-09-14 at 4 55 22 PM

gordonwatts:
Thanks for this bug!! Could you add a test case that breaks with 3 args but works with 2 before your fix goes in?

kyungeonchoi:
> Thanks for this bug!! Could you add a test case that breaks with 3 args but works with 2 before your fix goes in?

@gordonwatts - sorry for the slow response! I was on vacation. I added an `assert` to check 3rd arg is a string when `len(args) == 3`.

gordonwatts:
Very sorry - I totally missed this. I meant by adding a test case that there is a new `def test_xxx` function in the `tests` directory - if you look at the code coverage report here, it went down (hence the failing check by the CI). I'm trying to maintain the code coverage at something good here. :-)

servicex/servicex_utils.py
```diff
@@ -21,12 +21,15 @@ def _wrap_in_memory_sx_cache(fn):
...
-        assert len(args) == 2
+        assert len(args) == 2 or 3
         sx = args[0]
         from .servicex import ServiceXDataset
         assert isinstance(sx, ServiceXDataset)
         selection_query = args[1]
         assert isinstance(selection_query, str)
+        if len(args) == 3:
+            title = args[2]
+            assert isinstance(title, str)
```
gordonwatts, line 32: I do not think there is any need for these three lines, is there? The only key change is to line 24... And then a test that makes sure that a cached item with a title is fetched.

gordonwatts (changes requested):

servicex/servicex_utils.py
kyungeonchoi: Hi @gordonwatts - I realized that I've been totally forgotten this. 

Those three lines (L30-32) are just to confirm the type of argument for title is string. If it's redundant I can remove those lines :)

It seems like you already wrote tests for title - [L901 of `test_servicex.py`](https://github.com/ssl-hep/ServiceX_frontend/blob/30790674decaf140365fecb9cd9e6dce9ea49770/tests/test_servicex.py#L902). Do we need additional tests?

kyungeonchoi: Do you know how can we synchronize a title between frontend progress bar and dashboard entry?

servicex/servicex_utils.py
```diff
@@ -21,12 +21,15 @@ def _wrap_in_memory_sx_cache(fn):
     '''
     @functools.wraps(fn)
     async def cached_version_of_fn(*args, **kwargs):
-        assert len(args) == 2
+        assert len(args) == 2 or 3
```
AndrewEckart, line 24: Python does not interpret `len(args) == 2 or 3` the way you might think. It will read it as `(len(args) == 2) or (3)`. When there are 2 arguments, this expression evaluates to `True`, but when the number of arguments is not 2, the expression evaluates to `3` (the `or` operator simply returns the first truthy expression). This is not causing an error because `assert 3` is valid Python.

Should be changed to one of the options below to get the desired effect:
- `assert len(args) in (2, 3)`
- `assert 2 <= len(args) <= 3`
- `assert len(args) == 2 or len(args) == 3`

kyungeonchoi: @AndrewEckart - you are absolutely right! I will fix the bug.

gordonwatts (approved):

talvandaalen:
The fix is indeed covered by an existing test, so this is ready to merge.

PR #212, opened 2021-11-16
kafka removal

ivukotic:
simple change


servicex/cache.py
```diff
@@ -140,6 +138,11 @@ def __init__(self, cache_path: Path, ignore_cache: bool = False):
         self._path = cache_path
         self._ignore_cache = ignore_cache
 
+        # old windows versions have a limit of file path of 260 chars.
```
BenGalewsky, line 141: What does this have to do with Kafka?

ivukotic: it does not, but the test was failing so I fixed it.

BenGalewsky (approved):
Thanks for fixing the tests

Issue #215, opened 2021-12-21
Allow request result-dest to point to mounted posix volume


BenGalewsky:
_As a serviceX user I want my result files to be saved to a posix volume instead of object-store so I can use skyhook_

In order for ServiceX to write the result files to skyhook we used mounted POSIX volumes. This is a special helm chart config. The submitted requests require `result-destination` to be set to `volume`.

Currently, the value of `result-destination` is hardcoded

oshadura:
We have finally enabled Skyhook and currently testing its functionality @ UNL coffea-casa dev instance...

According docs https://github.com/ssl-hep/ServiceX/blob/develop/docs/deployment/output_options.md#posix-mounted-volumes,  I added in the ServiceX configuration:
```
    transformer:
      persistence:
        existingClaim: skyhook-pv-claim
        subdir: /mnt/cephfs/
```

BenGalewsky:
Until this issue is closed there is no way for you to tell ServiceX to write to the posix volume instead of the object store (which is hardcoded into the requests the Servicex_Frontend sends)

PR #217, opened 2022-01-31
Add optional constructor to change the result-destination

BenGalewsky:
## Problem
It is not possible to specify the `result-destination` for ServiceX Submits. It always defaults to `object-store`

Solution to #215 

## Approach
Added a new argument to the ServiceX constructor. Use this in the invocation of ServiceX Transform


gordonwatts:
Looks great and I think is ready to merge!

PR #231, opened 2022-03-23
Clean up

gordonwatts:
* Fix type errors (forgetting `Optional` in a number of places
* Fix spelling errors in comments
* start using black


gordonwatts:
* Fix type errors (forgetting `Optional` in a number of places
* Fix spelling errors in comments
* start using black

PR #235, opened 2022-04-27
Run with the new version of backoff

gordonwatts:
`backoff` bumpped to version 2.0

* We are using some of its internal code to emulate what we've got. This broke.


BenGalewsky:
Thank goodness!

gordonwatts:
Ha! I just discovered this by accident. How did you know?

BenGalewsky:
Was doing a clean install in a virtual environment and it required a pin to work

gordonwatts:
New version as soon as tests pass in CI...

Issue #238, opened 2022-05-02
URL_Stream methods should support S3 URIs


BenGalewsky:
# Problem
The newest versions of _coffea_ `NanoEvents` no longer support reading the parquet files with http protocol.

The correct way to handle this is to have the frontend return proper S3 URIs to the transformed files.

PR #239, opened 2022-05-02
Add as_signed_url property so url_streams can return s3 URIs by default

BenGalewsky:
# Problem
The latest version of _coffea_ `NanoEvents` can no longer read parquet files supplied via http.

Solution to #238 

# Approach
1. Changed the names of the get data url stream methods to be uri_stream to be more accurate
2. Added an optional parameter to the methods `as_signed_url` - by default this is False, but if you set it to True you can still get back the signed urls. Otherwise you get back s3 URIs


BenGalewsky:
Going to merge this now so we can get the Coffea build working again

PR #245, opened 2022-10-19
Added functions to get URLs sync from request + corresponding tests

talvandaalen:
* Implemented new functions to synchronously get all URLs from a request, for ROOT files and Parquet files. 
Functions are implemented as async to create the lists of URLs and then made sync after. Tests for each new function are also added.
* Fixed long-standing protection against `flake8` running during the initial stages (and got it back to passing flake8!!).

### DIFF
--- a/servicex/servicex.py
+++ b/servicex/servicex.py
@@ -68,7 +68,7 @@ class StreamInfoBase:
     def file(self) -> str:
         """Returns the ServiceX filename
 
-        This filename is unique in the dataset, and will be the same accross different queries
+        This filename is unique in the dataset, and will be the same across different queries
         against the dataset. It can be used as a key to sort results.
 
         Notes:
--- a/servicex/servicex.py
+++ b/servicex/servicex.py
@@ -471,9 +471,9 @@ class ServiceXDataset(ServiceXABC):
         selection_query: str,
         title: Optional[str] = None,
         as_signed_url: Optional[bool] = False,
-    ) -> list:
-        """Returns a list of each completed batch of work from ServiceX, taken from
-        the async iterator output of get_data_rootfiles_uri_stream.
+    ) -> List[StreamInfoUrl]:
+        """Returns a list of StreamInfoUrl entries containing a `url` for each output file
+        from the transform, taken from get_data_rootfiles_uri_stream.
         The data that comes back includes a `url` that can be accessed to download the
         data.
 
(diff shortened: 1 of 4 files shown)


servicex/servicex.py
```diff
@@ -465,6 +466,29 @@ async def get_data_rootfiles_uri_stream(
         ):  # type: ignore
             yield f_info
 
+    async def get_data_rootfiles_uri_async(
+        self,
+        selection_query: str,
+        title: Optional[str] = None,
+        as_signed_url: Optional[bool] = False,
+    ) -> list:
```
gordonwatts, line 474: The return type should be List[StreamInfoUrl].

gordonwatts, line 474: The return type should be List[StreamInfoUrl]

servicex/servicex.py
```diff
@@ -465,6 +466,29 @@ async def get_data_rootfiles_uri_stream(
...
+        """Returns a list of each completed batch of work from ServiceX, taken from
+        the async iterator output of get_data_rootfiles_uri_stream.
+        The data that comes back includes a `url` that can be accessed to download the
+        data.
```
gordonwatts, line 478: The data that comes back is a list of `StreamInfoUrl`'s that contain a url for each output file from the transform - or something similar.

servicex/servicex.py
```diff
@@ -483,7 +507,28 @@ async def get_data_parquet_uri_stream(
             selection_query, "parquet", title, as_signed_url
         ):  # type: ignore
             yield f_info
+            
+    async def get_data_parquet_uri_async(
+        self,
+        selection_query: str,
+        title: Optional[str] = None,
+        as_signed_url: Optional[bool] = False,
+    ) -> list:
```
gordonwatts, line 516: List[StreamInfoUrl]

servicex/servicex.py
```diff
@@ -483,7 +507,28 @@ async def get_data_parquet_uri_stream(
...
+        """Returns a list of each of the files from the minio bucket,
+        as the files are added there.
```
gordonwatts, line 518: Remove the "as the files are added there" - as this is going to wait until they are done and return all of them!

tests/test_servicex.py
```diff
@@ -797,6 +857,54 @@ async def test_stream_parquet_files_from_minio(mocker):
 
     assert mock_servicex_adaptor.query_json["result-format"] == "parquet"
 
+@pytest.mark.asyncio
```
gordonwatts, line 860: Have you run all your modified files through black? I'm seeing some differences in the formatting here which is making me think that didn't happen.

gordonwatts (changes requested):
Ok - There are only two real changes, and those are both type-hint related (so they don't cause any bugs except for in the editor). And a few comment updates. But looks fantastic! Update those changes, and use black on your modified files, and I think we are done!

gordonwatts (approved):

PR #247, opened 2022-10-19
ci: Update to stable GHA versions

matthewfeickert:
* Add workflow dispatch triggers to run CI on demand.
* Update actions to latest stable versions
* Fix CI so that `flake8` and `safety` checks actually run.
   - In the meantime this was fixed by @talvandaalen in PR #245. :+1: 
* Report coverage to Codecov for all the Python versions tested by separating uploads with version flags.

PR #248, opened 2022-10-19
feat: Add nox support

matthewfeickert:
Adopt the use of `nox` as a workflow runner and add a `noxfile.py` to provide `nox` "sessions" that allow for OS agnostic execution of scripted workflows. `nox` comes to me from introduction by @henryiii and I view it as much nicer replacement to `Makefile` driven tasks that makes it super simple to get people onboarded into a developer environment with just a `python -m pip install nox` and then to run the tests it is as easy as `nox -s tests`.

The reason I'm adding it here now is to make it very easy to build the docs across OSes when the docs workflow is added next (which will add `nox -s docs`).

* Update `.flake8` config.
   - Use 'extend-ignore' option to compliment command line ignores.
   - Use 'exclude' pattern to get most common files and tests.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,7 +27,7 @@ jobs:
           python -m pip list
       - name: Lint with Flake8
         run: |
-          flake8 --exclude=tests/* --ignore=E501,W503
+          flake8
       - name: Check for vulnerable libraries
         run: |
           python -m pip install safety
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,12 @@
 [flake8]
 max-line-length = 99
-ignore = W503
-exclude = .git,__pycache__,.nox,build,dist
+extend-ignore = E501,W503
+exclude =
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -27,7 +27,7 @@ jobs:
       - name: Lint with Flake8
         if: matrix.python-version == 3.7 && matrix.platform == 'ubuntu-latest'
         run: |
-          flake8 --exclude=tests/* --ignore=E501,W503
+          flake8
```
matthewfeickert, line 30: This will pass after PR #247 is merged.

matthewfeickert:
@gordonwatts This is ready for review but should follow PR #247.

gordonwatts:
How do you integrated `nox` into your `vscode` workflow?

gordonwatts:
I used the tests built-in integration so I can quickly debug a test or see the errors associated with that text inline with the editor.

PR #251, opened 2022-12-08
testing: try uproot5 / awkward2

lgray:
Spinning your CI to see if there's anything that flies off the rails with uproot5/awkward2.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,8 +40,8 @@ setup(
     install_requires=[
         "idna==2.10",  # Required to thread version needle with requests library
         "pandas~=1.0",
-        "uproot>=4.0.1, <5",
-        "awkward>=1.0.1, <2",
+        "uproot>=5.0.0rc7",
+        "awkward>=2.0.0rc6",
         "backoff>=2.0",
         "aiohttp~=3.6",
         "minio~=5.0",
--- a/setup.py
+++ b/setup.py
@@ -40,8 +40,8 @@ setup(
     install_requires=[
         "idna==2.10",  # Required to thread version needle with requests library
         "pandas~=1.0",
-        "uproot>=5.0.0rc7",
-        "awkward>=2.0.0rc6",
+        "uproot>=5",
+        "awkward>=2",
         "backoff>=2.0",
         "aiohttp~=3.6",
         "minio~=5.0",


lgray:
@gordonwatts please let this run :)

lgray:
@gordonwatts @talvandaalen can one of you approve this so the tests can run? It's possible you get this for free.

gordonwatts:
I don't see a way to "enable" this from the web (perhaps it has expired - sorry in that case!).

* I pulled everything from master (the 3.11 support and a bunch of improvements from @matthewfeickert have gone in).
* When I do a re-install, I see `Successfully installed awkward-2.0.6 awkward-cpp-7 servicex-0.1a1 uproot-5.0.2` get updated

Some tests do not pass any longer. Here is a stack dump for example:

```text
./tests/test_data_conversions.py::test_parquet_to_awkward Failed: [undefined]ModuleNotFoundError: No module named 'fsspec'
good_uproot_file_path = WindowsPath('tests/sample_uproot_servicex_output.parquet')

    @pytest.mark.asyncio
    async def test_parquet_to_awkward(good_uproot_file_path):
>       df = await DataConverterAdaptor("parquet").convert_to_awkward(good_uproot_file_path)

tests\test_data_conversions.py:53: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
servicex\data_conversions.py:54: in convert_to_awkward
    return await self._convert_parquet_to_awkward(file)
servicex\data_conversions.py:185: in _convert_parquet_to_awkward
    return await asyncio.wrap_future(_conversion_pool.submit(do_the_work, file))
..\..\..\AppData\Local\Programs\Python\Python311\Lib\concurrent\futures\thread.py:58: in run
    result = self.fn(*self.args, **self.kwargs)
servicex\data_conversions.py:183: in do_the_work
    return ak.from_parquet(str(file))  # type: ignore
.venv\Lib\site-packages\awkward\operations\ak_from_parquet.py:69: in from_parquet
    parquet_columns, subform, actual_paths, fs, subrg, row_counts, meta = metadata(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

path = 'tests\\sample_uproot_servicex_output.parquet', storage_options = None
row_groups = None, columns = None, ignore_metadata = False, scan_files = True

    def metadata(
        path,
        storage_options=None,
        row_groups=None,
        columns=None,
        ignore_metadata=False,
        scan_files=True,
    ):
        import awkward._connect.pyarrow
    
        # early exit if missing deps
        pyarrow_parquet = awkward._connect.pyarrow.import_pyarrow_parquet("ak.from_parquet")
>       import fsspec.parquet
E       ModuleNotFoundError: No module named 'fsspec'

.venv\Lib\site-packages\awkward\operations\ak_from_parquet.py:102: ModuleNotFoundError
```

The other thing is `uproot`'s `lazy` is used.

I think if both of those are fixed, then all tests will be green again. Can you suggest what should be done in each case? The `fsspec` should probably be a fix in `awkward` (??), and the `laxy` I suspect @masonproffitt  can tell me what to replace it with.

matthewfeickert:
> And I suppose when fixing `laxy` there isn't a real way to be compatible with both versions of awkward? What should we do to make sure both are supported? Or should we start requiring `awkward2` everywhere now. :-(

Yeah, `ak.lazy` doesn't exist in `uproot` `v5` as it has been replaced with `dask-awkward`

c.f. https://github.com/scikit-hep/uproot5/discussions/796, in particular

> ... the major part of which was replacing uproot.lazy, which is based on Awkward 1's virtual and partitioned lazy arrays, with the new Dask collection, [dask-awkward](https://github.com/dask-contrib/dask-awkward). The entry point for this function is uproot.dask.

ponyisi:
I assume we can close

PR #271, opened 2023-07-11
Safe Downloaded Filenames

gordonwatts:
Make sure that filenames for files coming back from minio have ":", "*", etc., removed so they are safe on any OS


BenGalewsky (approved):
yes

PR #272, opened 2023-07-11
Robust sync async routines

gordonwatts:
* Use the library `make_it_sync` to support various environments (like command line, jupyter, etc.).


BenGalewsky (approved):
yep

PR #273, opened 2023-07-12
Short Filenames

gordonwatts:
* Fixed bug in short-filename logic
* The max filename should never be shorter than the hash length, or the result is undefiend.
* Changed filename length to be 60.


BenGalewsky (approved):
yow

PR #298, opened 2023-07-18
Accept both .servicex and servicex.yaml config files

talvandaalen:
Small changes to configuration.py to search for both .servicex and servicex.yaml config files, while also accepting a user-defined config file path.


BenGalewsky (approved):
Ok

PR #300, opened 2023-07-23
Handle missing cache_dir and expand bengal1 in path

### LINKED ISSUE #297: Undefined cache_path gives unclear error
If the cache_path is undefined in the .servicex file it gives a Nonetype error for a path being parsed, but it should just tell the user to set a cache_path variable in their config.

The cache_path being undefined can happen easily if a user simply copy-pastes a config from their backend profile, since that doesn't contain a cache_path variable yet.

BenGalewsky:
# Problem
There is no default for cache_path if not provided in the .servicex file. The error messages are unclear as well.

We also want to support expanding ${USER} in the path to make it easy to support multiple users on a single JupyterHub node

Fixes #297

### DIFF
--- a/servicex/configuration.py
+++ b/servicex/configuration.py
@@ -26,10 +26,11 @@
 # OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 import os
-from pathlib import Path
+import tempfile
+from pathlib import Path, PurePath
 from typing import List, Optional, Dict
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, validator
 
 import yaml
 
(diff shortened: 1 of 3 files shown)


servicex/configuration.py
```diff
@@ -46,6 +47,35 @@ class Configuration(BaseModel):
...
+    @validator("cache_path", always=True)
+    def expand_cache_path(cls, v):
+        """
+        Expand the cache path to a full path, and create it if it doesn't exist.
+        Expand ${USER} to be the user name on the system. Works for windows, too.
+        :param v:
+        :return:
+        """
+        if not v:
+            v = "/tmp"
```
gordonwatts, line 59: Two questions about this line:
* does this work on windows?
* /tmp is machine wide, and you probably need something that is per-user. So you should do /tmp/<user> I think. At least, I have that code in the original and I had to do that for a reason... I think it had to do with wanting to clear the cache, and only clearing my stuff and not others that were on the same machine.

gordonwatts (commented):
Looks great other than the two quick questions.

servicex/configuration.py
```diff
@@ -46,6 +47,35 @@ class Configuration(BaseModel):
...
+
+        # If they have tried to use the USER or UserName as an expansion, and it has failed, then
+        # translate it to maintain harmony across platforms.
+        if "${USER}" in s_path and "UserName" in os.environ:
+            s_path = s_path.replace("${USER}", os.environ["UserName"])
+        if "${USER}" in s_path and "USER" in os.environ:
+            s_path = s_path.replace("${USER}", os.environ["USER"])
+
+        p_p = PurePath(s_path)
+        if len(p_p.parts) > 1 and p_p.parts[1] == "tmp":
```
BenGalewsky, line 71: @gordonwatts this line replaces `/tmp` with the correct temporary directory for the platform.

Issue #283, opened 2023-07-14
Progress Bar Updates


gordonwatts:
* Must be able to run simultanious transforms in Jupyter
* Make it possible to turn off the progress bars

PR #301, opened 2023-07-28
Dataset groups Revisted

### LINKED ISSUE #278: DatasetGroup Allows for Applying Select and Where Statements Across Multiple Datasets
_As an analyzer, I want to collect a series of datasets and apply func_adl and serviceX operations across them all so I can treat them as a single logical dataset_

# Assumptions
1. Constructor takes a list of func_adl datasets
2. Provides covers for `Select`, `SelectMany` and `Where` operations that return the DatasetGroup
3. Provides covers for `as_files` and `as_signed_urls` that launch the operations and gather the results

BenGalewsky:
# Problem
The first attempt at a dataset group was a bit short sighted. It should accept datasets and then provide cover methods for the "verbs" (as_files, as_pandas). Also the progress bars are cute, but don't allow for multiple concurrent jobs since they all need to share a common Progress context.

Fixes #278 and #283 

# Approach
Fixed the interface for dataset_group to accept a list of datasets. The cover methods were pretty simple at that point.

The progress bars required a rethink. The verb methods all accept a `display_progress` argument that disables this completely since some people may not like them regardless of how pretty they are...

Wrote a new context manager, `ExpandableProgress` - this is allows for an existing, `start`ed progress bar to be injected into the submit workflow. It handles nesting so that you can pass an instance of `ExpandableProgress` into a verb method and it will just pull out the progress bar and carry on.

Also added an example using DatasetGroup. This is cribbed from the joiner notebook, but I skipped the metadata so we don't need extra libraries to run the examples....


servicex/dataset_group.py
```diff
@@ -26,29 +26,47 @@
...
+    ) -> List[TransformedResults]:
+        with ExpandableProgress(display_progress, provided_progress) as progress:
+            self.tasks = [
+                d.as_signed_urls_async(provided_progress=progress)
+                for d in self.datasets
+            ]
+            return await asyncio.gather(*self.tasks)
 
-    gather_results = make_sync(gather_results_async)
+    as_signed_urls = make_sync(as_signed_urls_async)
```
gordonwatts, line 72: What happened to `Select` and `Where`, etc.?

servicex/expandable_progress.py
```diff
@@ -0,0 +1,92 @@
...
+from typing import Optional
+
+from rich.progress import Progress, TextColumn, BarColumn, MofNCompleteColumn, \
+    TimeRemainingColumn
+
+
+class ExpandableProgress:
+    def __init__(self,
+                 display_progress: bool = True,
+                 provided_progress: Optional[Progress | ExpandableProgress] = None):
```
gordonwatts, line 39: There are a bunch of other places in the API where you just allow Progress - but it looks like you need to always allow `Progress | ExpandableProgress` (since they aren't subclasses). Perhaps define a new type `ProgressIndicators = Progress | ExpandableProgress` and wherever you have `Progress` replace?

servicex/dataset_group.py
```diff
@@ -26,29 +26,47 @@
...
+
+from servicex.dataset import Dataset
+from servicex.expandable_progress import ExpandableProgress
+from servicex.models import TransformedResults, ResultFormat
 from make_it_sync import make_sync
 
 
 class DatasetGroup:
-    def __init__(self, datasets: List[Coroutine]):
+    def __init__(self, datasets: List[Dataset]):
```
gordonwatts, line 40: Should this just be a DataSet - so you can have a dataset group as part of a dataset group?

gordonwatts:
Left a few comments - looks good. Some deisgn questions...

servicex/dataset_group.py
```diff
@@ -26,29 +26,59 @@
...
+from servicex.dataset import Dataset
+from servicex.expandable_progress import ExpandableProgress
+from servicex.func_adl.func_adl_dataset import FuncADLDataset
+from servicex.models import TransformedResults, ResultFormat
 from make_it_sync import make_sync
 
 
 class DatasetGroup:
-    def __init__(self, datasets: List[Coroutine]):
+    def __init__(self, datasets: List[Union[Dataset, FuncADLDataset]]):
```
gordonwatts, line 41: Could we also include `DatasetGroup` in the list of things here?

servicex/dataset_group.py
gordonwatts: These are now in the func adl dataset group. Cool.

servicex/dataset_group.py
gordonwatts: And now shouldn't this be something from types.py?

servicex/expandable_progress.py
BenGalewsky: Good idea - users no longer can provide a bare `rich.Progress`. The public API only allows `ExpandableProgress`  - I did make use of the type to make the function signatures more concise.

servicex/dataset_group.py
BenGalewsky: Fixed

servicex/types.py
```diff
@@ -27,6 +27,11 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 from typing import Union
 
+from rich.progress import Progress
```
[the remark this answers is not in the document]
BenGalewsky: Moved this out of types.py

PR #305, opened 2023-09-06
Add a cache for the code generators

prajwalkkumar:
This commit adds functionality to cache the code generators for offline processing.

### DIFF
--- a/servicex/app/codegen.py
+++ b/servicex/app/codegen.py
@@ -25,41 +25,36 @@
 # CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
 # OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-from typing import Optional
 
 import rich
 import typer
 
-from servicex._version import __version__
-from servicex.app.transforms import transforms_app
-from servicex.app.cache import cache_app
-from servicex.app.codegen import codegen_app
+from servicex.app.cli_options import url_cli_option, backend_cli_option
+from servicex.servicex_client import ServiceXClient
+from typing import Optional
 
-app = typer.Typer(no_args_is_help=True)
-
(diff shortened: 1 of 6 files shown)


servicex/app/codegen.py
```diff
@@ -0,0 +1,56 @@
...
+        url: Optional[str] = url_cli_option,
+        backend: Optional[str] = backend_cli_option,
+        flush: Optional[bool] = typer.Option(None,
+                                             "--flush",
+                                             help="Flush the codegens from the cache"
+                                             )):
+    """
+    List the available code generators
+    """
+    sx = ServiceXClient(url=url, backend=backend)
```
BenGalewsky, line 50: This should use the cache! I tried switching off wifi and this call didn't work

servicex/app/codegen.py
```diff
@@ -0,0 +1,56 @@
...
+
+from servicex.app.cli_options import url_cli_option, backend_cli_option
+from servicex.servicex_client import ServiceXClient
+from typing import Optional
+
+codegen_app = typer.Typer(name="codegen", no_args_is_help=True)
+
+
+@codegen_app.command(no_args_is_help=False)
+def codegens(
```
BenGalewsky, line 40: This winds up with a weird command line
```
servicex codegen codegens --flush
```

Maybe 
```
servicex codegens flush
and 
servicex codegens list
```

BenGalewsky (changes requested):
Weird CLI user experience - `servicex codegen codegens` doesn't make sense
I tried 
`servicex codegen codegens --flush`
then
` servicex codegen codegens` prints out 
```
Returning code generators from cache
```
They should not be in the cache at that point!

The examples are now working - try these out and make sure this all works as expected. Turn off your wifi and make sure you can get results out of cache if you are on an airplane

servicex/app/codegen.py
prajwalkkumar: This was using the cache. It works now with wifi off after some more changes.

servicex/app/codegen.py
prajwalkkumar: This has the expected cli now. 
I moved it out of main.py and hence we were seeing this behaviour. 
servicex codegens flush
servicex codegens list work now

prajwalkkumar:
Made the changes requested!

PR #306, opened 2023-09-11
Modify to use logger instead of rich prints

prajwalkkumar:


servicex/dataset.py
```diff
@@ -59,6 +61,14 @@
 
 from make_it_sync import make_sync
 
+FORMAT = "%(message)s"
+logging.basicConfig(
+    format=FORMAT,
+    datefmt="[%X]",
+    handlers=[RichHandler(markup=True,
+                          rich_tracebacks=True)],
+)
```
alexander-held, line 70: As far as I remember, having this called automatically means that users of this library cannot externally configure logging. This can only really be called once and now it looks like that would happen automatically upon `servicex` import.

To me, use of `logging` implies giving users control (appropriate for a library). A sensible default setting like this could be moved to some dedicated function living in `servicex` that users can manually call if desired. Having this all set automatically without possibility to customize would be more appropriate for an application.

prajwalkkumar: Makes, sense. I have now moved it to a separate file and initialized a rich_logger, that is used to log logs that need to be made pretty using the rich handler. This does not call logging.basicConfig anymore, and that is reserved for use by the application and not the library. Can you please review this PR and see if it makes sense now?

alexander-held: I am not very familiar with running logging through `rich`, but currently it does not work how I would expect it to. If I use
```python
from servicex import servicex_logger
servicex_logger.init_logger()
```
I get INFO-level messages (as expected).

If I however want to customize my logger via e.g.
```python
import logging
logging.getLogger("servicex").setLevel(logging.INFO)
```
I do not get those INFO-level messages. Confusingly,
```python
logging.getLogger().setLevel(logging.INFO)
```
does show me the INFO-level messages from ServiceX (amongst others), so am I using the wrong package name?

PR #314, opened 2023-09-22
Deal with empty and bad dataset names

### LINKED ISSUE #312: Make sure we can handle an empty dataset properly
See [this bug](https://github.com/ssl-hep/ServiceX/issues/640) in the `ServiceX` project for context.

### LINKED ISSUE #313: Bad Dataset Name to Rucio and the query hangs
See this [sample](https://github.com/gordonwatts/func-adl-servicex-demos/blob/main/xAOD/03-bad-dataset.ipynb) for an example of a query that causes a hang. No error, no timeout with zero files, nothing. No detection the run is complete.

gordonwatts:
ServiceX no longer returns a *Fatal* error when it is asked to look up a bad dataset name or a dataset name that has zero files.

* Detect a completed transform with no files

Fixes #313
Fixes #312

### DIFF
--- a/servicex/servicex_adaptor.py
+++ b/servicex/servicex_adaptor.py
@@ -210,6 +210,18 @@ class ServiceXAdaptor:
                     f"Transform status for {request_id}" ' is marked "Fatal".'
                 )
 
+            if (
+                "status" in info
+                and info["status"] == "Complete"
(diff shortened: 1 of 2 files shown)


BenGalewsky (approved):
This seems good to have anyway...

PR #315, opened 2023-09-22
Agc fixes

BenGalewsky:
As I was testing out the ServiceX client 3.0 with the Analysis grand Challenge notebook, I uncovered a few minor issues.

1.  DatasetGroup didn't return self in set_result_format so this couldn't be used in a pipeline
2. The import statement for types in func_adl_dataset didn't use the correct module name
3. Added ignore_cache option

### DIFF
--- a/examples/xAODDataset.py
+++ b/examples/xAODDataset.py
@@ -47,9 +47,8 @@ good_ele = ds.Select(
     lambda e: {
         "run": e.EventInfo("EventInfo").runNumber(),
         "event": e.EventInfo("EventInfo").eventNumber(),
-        "good_ele": e.Electrons("Electrons").
-        Where(lambda e: (e.pt() / 1000 > 25.0) and (abs(e.eta()) < 2.5)
-              ),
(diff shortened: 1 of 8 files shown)


examples/dataset_group.py
```diff
@@ -1,4 +1,7 @@
 from servicex import ServiceXClient, RucioDatasetIdentifier, ResultFormat, DatasetGroup
+from rich.console import Console
```
gordonwatts, line 2: ALl this looks great - other than we aren't using `logging` - but how did this work before - I see "Console" used below and it doesn't show up in this diff! :-)

gordonwatts: Oh - I see - the include was down below. :-)

gordonwatts (approved):
Does the usual `with ignore_cache():` work? There is a bunch of code out there that supports that and people have gotten quite used to it.

examples/dataset_group.py
BenGalewsky: yeah - just a flake8 issue about my imports lower in the file

PR #317, opened 2023-09-22
Awkward 2.0 support in `servicex` 2.0

### LINKED ISSUE #316: Support awkward 2.0
Add support for awkward 2.0 rather than just 1.0.

gordonwatts:
Enable awkward 2.0 support in `servicex`

* Setup configured to support only awkward 1.0 if you are on python 3.7 (we need to drop support for this!)
* If you load a `awkward` array using this library, it will give you an `awkward_dask` array if `awkward` 2.0 is loaded, or a `lazy` array if `awkward` 1.0 is in your environment.

Fixes #316


gordonwatts:
This is a stop-gap fix until 3.0 is released, so I'm not too worried about the drop in code coverage.

BenGalewsky (approved):
Glad we can evolve

PR #318, opened 2023-09-25
Don't used cached results if there were any failures

BenGalewsky:
# Problem
Even if there were failures in a previous transform we would just pull it out of cache and serve it up. 

# Approach

Keep track of the number of failed files in the cached record. Only return it to the user if it was flawless.

1. As part of this PR I realized that the tiny DB is not threadsafe so I added locks to the cache code.
2. Added reporting of failed files in the cache CLI report
3. There was a bug `servicex cache list` command where it would try to report on code generator records in the tiny db. Enhanced that query to only report transform records

### DIFF
--- a/servicex/app/cache.py
+++ b/servicex/app/cache.py
@@ -57,7 +57,7 @@ def list():
     table.add_column("Codegen")
     table.add_column("Transform ID")
     table.add_column("Run Date")
-    table.add_column("Files")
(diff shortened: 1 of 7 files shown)


kyungeonchoi:
Save only 3

BenGalewsky:
This has been fixed in other PRs now

PR #324, opened 2023-10-09
Match the root file result format in the generated json to the expected value in the service

BenGalewsky:
For consistency sake, the enum was changed to root, however this is not what the server expects. To fix this we keep the enum key root, but make the string representation root-file. Added a unit test to verify this along with a pointer to the statement in the ServiceX client which sets up the valid values for this field

### DIFF
--- a/servicex/models.py
+++ b/servicex/models.py
@@ -46,7 +46,7 @@ class ResultFormat(str, Enum):
     Specify the file format for the generated output
     """
     parquet = "parquet"
-    root = "root"
+    root = "root-file"
 
 
 class Status(str, Enum):
--- a/tests/test_servicex_adapter.py
+++ b/tests/test_servicex_adapter.py
@@ -43,6 +43,16 @@ def servicex():
     return ServiceXAdapter("https://servicex.org")
 
 
+def test_result_formats():
+    """
+    This test is just to make sure the enum string representations match the values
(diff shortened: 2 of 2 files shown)


alexander-held (commented):
The following setup breaks for me with this update:
```python
from servicex.dataset_identifier import FileListDataset
from servicex.servicex_client import ServiceXClient

sx = ServiceXClient(backend="uproot")
dataset_id = FileListDataset("root://eospublic.cern.ch//eos/opendata/atlas/OutreachDatasets/2020-01-22/4lep/MC/mc_345060.ggH125_ZZ4lep.4lep.root")
ds = (sx
      .func_adl_dataset(dataset_id, codegen="uproot", title="Root", result_format="root")
      .set_tree("mini")
)
files_root = ds.Select(lambda e: {'lep_pt': e['lep_pt']}).Where(lambda e: e['lep_pt'] > 1000).as_files()
```
via
```pytb
pydantic.error_wrappers.ValidationError: 1 validation error for TransformRequest
result-format
  value is not a valid enumeration member; permitted: 'parquet', 'root-file' (type=type_error.enum; enum_values=[<ResultFormat.parquet: 'parquet'>, <ResultFormat.root: 'root-file'>])
```
`result_format=ResultFormat.root` works.

PR #333, opened 2023-12-05
First DataBinder integration

kyungeonchoi:
- First ServiceX DataBinder integration
- Only for Python transformer for this PR


servicex/databinder/databinder.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+
+    """
+    def __init__(self, config: Union[str, Path, Dict[str, Any]]):
+        self._config = load_databinder_config(config)
+        self._sx_ds = DataBinderDeliver(self._config)
+
+    def deliver(self, overall_progress_only: bool = False):
+        out_paths_dict = asyncio.run(self._sx_ds.get_data())
+
```
BenGalewsky, line 46: Don't check in commented out code. Please delete these lines if they aren't needed

servicex/databinder/databinder_configuration.py
```diff
@@ -0,0 +1,257 @@
...
+        if 'IgnoreLocalCache' not in sample.keys():
+            if 'IgnoreLocalCache' in config['General'].keys():
+                config['Sample'][idx]['IgnoreLocalCache'] = config['General']['IgnoreLocalCache']
+            else:
+                config['Sample'][idx]['IgnoreLocalCache'] = False
+
+    return config
+
+
+def _replace_definition_in_sample_block(config: Dict[str, Any]):
```
BenGalewsky, line 90: Wow! What is this method doing? Can you write up a comment describing what is going on here?

servicex/databinder/databinder_configuration.py
```diff
@@ -0,0 +1,257 @@
...
+            elif sample['Codegen'] == "python":
+                config['Sample'][idx]['Transformer'] = "uproot"
+        else:
+            config['Sample'][idx]['Transformer'] = pair[0]
+            config['Sample'][idx]['Codegen'] = pair[1]
+
+    return config
+
+
+def _validate_config(config: Dict[str, Any]):
```
BenGalewsky, line 164: It would be so much simpler to replace this generic dict with a [pydantic](https://docs.pydantic.dev/latest/) model (see models.py)

servicex/databinder/databinder_requests.py
```diff
@@ -0,0 +1,148 @@
...
+                query = sample['Function']
+                return _get_servicex_dataset(sample).with_uproot_function(
+                    query
+                )
+            else:
+                raise TypeError(
+                    f"Unknown code-generator in Sample {sample['Name']}"
+                )
+
+        # def _servicex_dataset_query_output(sample):
```
BenGalewsky, line 128: Delete commented out code

BenGalewsky (changes requested):
This is a great change for the client!
I pointed out some commented-out code. We generally don't want that in the final PR, so please delete those lines if they aren't needed.

A couple of methods are very complicated and deserve comments explaining them.

I pushed a [commit to the ssl-hep repo](https://github.com/ssl-hep/ServiceX_frontend/commit/236331ca9d9bfe351ea1a51bad702ccf0bf89cdf) with a new test for the string-based python functions and also to correct the method type signature to match this new usage. I don't have write permission to your repo to add this commit to the PR myself. You should be able to cherry pick this commit

I pointed out that this code would be much simpler if the databinder config was a [pydantic](https://docs.pydantic.dev/latest/) class instead of a raw dict. It's ok if you want to tackle that in a subsequent PR and not hold this up.

BenGalewsky:
I see that you have write permission to this repo. Instead of a fork, your could move your changes to a branch of this repo so I can also push commits to it.

BenGalewsky (approved):
Ok - thanks for the cleanup

PR #339, opened 2024-01-24
Update examples for servicex backend v1.3.2

kyungeonchoi:
Update examples for servicex backend v1.3.2


BenGalewsky (approved):
Thanks for cleaning up these examples

PR #341, opened 2024-02-29
None value check for Definition key 

ketan96-m:
### Problem:
In the `ServiceXSpec` model in `databinder_models.py`, the function `replace_definition` has a check to see if the `Definition` key exists in the `values` dictionary. However, running the bigger_uproot.py I noticed that key had a value None and this was not checked and was throwing an AttributeError.

### Fix:
Added another condition to check for None value for the key `Definition`.


BenGalewsky (approved):
Nice PR! Thank you

PR #342, opened 2024-03-07
Changes in the cosmetics of databinder integration

kyungeonchoi:
- Mostly clear from the commit messages
- Changes enable the notebook example for ACAT 2024 poster


gordonwatts:
Nice. Is there some example code of how this would work? We really need docs in the code itself.

PR #343, opened 2024-03-11
Cache Clear fix on Temp files not deleting

ketan96-m:
### Problem:
When user does not give `cache_path` in `.servicex`, the default temp directory is used. Therefore when you try to clear the cache, some files are not deleted due to permission issue.

### Fix:
Rather than having `/tmp` as the default added `/tmp/servicex_$USER`. 
This ensures the local cache is set to this path and it is safely deleted when clearing cache

### Code changes:
In the configuration.py-> `expand_cache_path(cls, v)`, changed the default value of `v="/tmp"` to `v="/tmp/servicex_$USER"`

Added tests for this new changes in test_config.py

### DIFF
--- a/servicex/configuration.py
+++ b/servicex/configuration.py
@@ -55,7 +55,7 @@ class Configuration(BaseModel):
         :param v:
         :return:
         """
-        #create a folder inside the tmp directory if not specified in cache_path
+        # create a folder inside the tmp directory if not specified in cache_path
         if not v:
             v = "/tmp/servicex_${USER}"
 
--- a/servicex/configuration.py
+++ b/servicex/configuration.py
@@ -55,8 +55,9 @@ class Configuration(BaseModel):
         :param v:
         :return:
         """
+        #create a folder inside the tmp directory if not specified in cache_path
         if not v:
-            v = "/tmp"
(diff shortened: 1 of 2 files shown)


BenGalewsky (changes requested):
This looks great. I didn't know about the ${USER} interpolation and can't find it anywhere in a google search. I see that the test passes, so it must work, but it's a little surprising!
Can you do an [interactive rebase](https://medium.com/nona-web/cleaning-up-commit-history-with-git-rebase-61dc41d2829c) to eliminate your extra commits

ketan96-m:
The reason I used the ${USER} because inside the same function, a condition was present to replace this with the username of the system.
```
if "${USER}" in s_path and "UserName" in os.environ:
            s_path = s_path.replace("${USER}", os.environ["UserName"])
```

ponyisi:
Suggestion: if we keep the ${USER} interpolation (which I'm fairly neutral on) I suggest that instead of the environment variable logic, we switch to the standard library `os.getlogin` function (otherwise we are hard-coding assumptions about the client systems).

BenGalewsky (approved):
Thanks. this looks good and avoids an unexpected conflict in the tmp dir

PR #348, opened 2024-03-22
Remove poetry lock, don't use implicit upper bounds on versions, fix Py 3.12 compatibility

ponyisi:
Also fix a couple of things flake8 caught.

Ugly history - hope it can get squashed if accepted.


ponyisi:
Obsoleted by #349 and followups

PR #359, opened 2024-04-17
Multiple xrootdfiles

BenGalewsky:
# Problem
The Pydantic 3.0 client doesn't accept multiple root file paths. It only supports a single root file. 

# Approach
Changed the name of the property from `XRootdFile` to `XRootDFiles` and allow it to be a string or a list of strings. Note that the case changed on this too.

Need to start testing more of this functionality out, so I updated test_databinder as a start


kyungeonchoi (approved):
Fixed typos :)

Issue #363, opened 2024-04-29
Update pydantic to v2 API


matthewfeickert:
At the moment on the `3.0_develop` branch, the `pydantic` requirement is constrained to

https://github.com/ssl-hep/ServiceX_frontend/blob/a7f11f50e0074267fdb9f8e35d85b12103d2a97f/pyproject.toml#L38)

(as that branch isn't stable forever, for posterity

```toml
dependencies = [
    "qastle>=0.17",
    "func_adl>=3.2.6",
    "requests>=2.31",
    "pydantic>=1.10,<2.0",  # FIXME: Adopt pydantic 2.0 API
    ...
]
```

)

In https://github.com/iris-hep/idap-200gbps-atlas/issues/41 @gordonwatts had identified this as a blocker for using in the same stack as other tools like `correctionlib`. Updating `pydantic` to use `v2` would be highly impactful.

matthewfeickert:
cc @kyungeonchoi and @BenGalewsky as they are probably the relevant people to discuss this with beyond / in addition to @gordonwatts.

matthewfeickert:
I tried to give this a go on my fork's `feat/update-to-pydantic-v2-api` branch, but didn't get super far as I'm inexperienced with `pydantic` and haven't yet worked through the [SciPy 2023 `pydantic` tutorial](https://github.com/adonath/scipy-2023-pydantic-tutorial) to teach myself. Example, it took me over 30 minutes to understand that

https://github.com/ssl-hep/ServiceX_frontend/blob/a7f11f50e0074267fdb9f8e35d85b12103d2a97f/servicex/databinder_models.py#L121

(Posterity repost as the `3.0_develop` branch will not live forever
```python
class ServiceXSpec(BaseModel):
    General: General
    Sample: List[Sample]
    Definition: Optional[Definition]
```
)

is getting hit by the field having the same name as a model (https://github.com/pydantic/pydantic/issues/7871#issuecomment-1770958931) and so you need to change the field name

```python
...
    definition: Optional[Definition]
```

.

@BenGalewsky as you were the one to move things over to `pydantic` in PR https://github.com/ssl-hep/ServiceX_frontend/pull/340 you might be in the best position to either give advice if you have the time to scout the differences, or also do the implementation changes if they all appear trivial to you.

matthewfeickert:
Resolved by PR #368

PR #364, opened 2024-05-01
build: Update typer lower bound to v0.12.1

matthewfeickert:
This is a component PR on the way towards targeting Issue #363, which is needed for full integration in the IDAP pipelines.

* In typer v0.12.1 the 'all' extra is dropped as all required dependencies are provided as typer installs the 'standard' extra by default.
   - c.f. https://github.com/tiangolo/typer/releases/tag/0.12.1
* The increase in version number only increases the lower bound on the click dependency and removes multiple upper bounds on other dependencies, making it easier to support the newer release.
   - c.f. https://github.com/tiangolo/typer/compare/0.9.1...0.12.1


matthewfeickert:
@gordonwatts @BenGalewsky @ponyisi This is ready for review.

BenGalewsky (approved):
Sensible improvement. Thank you @matthewfeickert

PR #365, opened 2024-05-01
ci: Use uv for all pip installs

matthewfeickert:
* Use 'uv pip' for all calls to 'pip install' in GitHub Actions workflows.
   - c.f. https://github.com/astral-sh/uv/
   - In the case of the serviceX client, this is just a few seconds saved, but that adds up over every CI run.
* The '--system' call is required as uv requires a virtual environment to be activated without it, and GitHub Actions does not provide a default virtual environment.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -22,7 +22,8 @@ jobs:
 
       - name: Lint with Flake8
         run: |
-          python -m pip install --upgrade flake8
+          python -m pip install --upgrade uv
+          uv pip install --system --upgrade flake8
           flake8
 
   test:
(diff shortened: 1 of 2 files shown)


matthewfeickert:
cc @BenGalewsky for review.

BenGalewsky (approved):
If you recommend this, then I'm trying it out!

matthewfeickert:
> If you recommend this, then I'm trying it out!

It can have significant time savings, as https://github.com/scikit-hep/pyhf/pull/2444#issuecomment-2081222540 isn't even the largest install time drops that I've seen with it. Henry has a bunch more examples, and thanks to him it is also available as an installer option for `build` (c.f. https://github.com/pypa/build/releases/tag/1.2.0).

PR #369, opened 2024-05-02
Allow timeout for transform submission

gordonwatts:
The default transform submission timeout is 5 seconds. In cases where ServiceX has a very large cached filelist (say 65K!) it can take more than 5 seconds for it to reply.

This change makes the client more robust against that sitaution.

## Minor

* Update old vscode settings

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -102,5 +102,5 @@
     ],
     "python.testing.unittestEnabled": false,
     "python.testing.pytestEnabled": true,
-    "editor.formatOnSave": true,
+    "editor.formatOnSave": false,
 }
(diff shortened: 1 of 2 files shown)


gordonwatts (commented):
It is likely the reason is behind these very long weights is the amount of time it takes to load in a very large list of files (65K) from the cache.

See https://github.com/iris-hep/idap-200gbps-atlas/issues/84 for more info.

BenGalewsky:
Actually, I'm not so sure it's polite to check vscode settings into the repo, but since they were already there...

BenGalewsky (approved):
Actually, I'm not so sure it's polite to check vscode settings into the repo, but since they were already there...

PR #371, opened 2024-05-03
Use logging library for status updates

gordonwatts:
This PR will exchange some of the `rich.print` calls that happen as the library runs and route them through `logging.info` and `logging.warning` and `logging.error`.

* Less inline output while library is running
* Allows user of library to control how much output is printed
* Brings it into line for the "correct" way to use library, according to python.

### DIFF
--- a/servicex/query.py
+++ b/servicex/query.py
@@ -43,7 +43,6 @@ except ModuleNotFoundError:
 
 from servicex.types import DID
 
-import rich
 from rich.progress import Progress, TaskID
 
 from servicex.configuration import Configuration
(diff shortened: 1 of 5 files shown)


gordonwatts (commented):
There is a chance that some of these `logging.info` should be downgraded to `logging.debug`.

servicex/query.py
```diff
@@ -157,9 +156,10 @@ def set_result_format(self, result_format: ResultFormat):
         return self
 
     async def submit_and_download(
-        self, signed_urls_only: bool,
```
BenGalewsky, line 160: Formatting changes in the same commit as the code changes make the review a lot harder. If you want to change formatting, it's more polite to make that a single commit and then have the meat of the PR in a second commit

BenGalewsky (changes requested):
Could you make the logger class variables in these two classes? It seems messy to keep repeating the `logging.get_logger("__name__")`

PR #372, opened 2024-05-07
Handle unknown status code being returned in submission

ponyisi:
Be more informative if e.g. status 500 is returned from ServiceX


BenGalewsky (approved):
Thank you

PR #377, opened 2024-05-09
3.0 Client Documentation

ketan96-m:
Created .rst file structure for documenting the client library

It has the following outline:
1. Introduction
2. Installation
3. Getting Started
4. Tranformer Matrix
5. DataBinder
6. Examples
7. ServiceX Contributor Guide
8. Troubleshooting
9. About Page
10. Sample
11. ServiceX api docs

Using sphinx documentation to build the .html files
commands to make the api docs: 
`sphinx-apidoc -f -o docs/ servicex/`

Build html
`make html`

Clean html
`make clean html`

More details on sphinx: https://www.sphinx-doc.org/en/master/tutorial/first-steps.html#building-your-html-documentation


ketan96-m:
I have only included one bigger uproot example using code-include extension.
The bigger uproot example file had to be refactored into a method to avoid running the example whenever I build the html.
code-include extension uses the directive 
`.. code-include :: :func:"examples.bigger_uproot.bigger_uproot"`

gordonwatts:
I like this a lot! I think it is important to get this in there, and then we start fixing and modifying it as we go. A few things we should have in future PR's:

* Build this as part of the CI
* Build and deploy it on tags or similar so that the docs are kept up to date in a place we can point people to - and have it done automatically.

BenGalewsky:
I'm having trouble building the docs:
1. Need to `cd docs` before issuing the`make html` command
2. Need to `pip install .[docs]`
3. Tried `make html` and I get this error:
```
Running Sphinx v7.2.6

Extension error:
Could not import extension code_include.extension (exception: No module named 'code_include')
make: *** [html] Error 2
```

ponyisi:
Hi @BenGalewsky - does it build for you now?

PR #378, opened 2024-05-12
Properly display error from transform

gordonwatts:
If ServiceX responds with a > 400 error on the transform WebAPI, properly display the error message.


BenGalewsky (approved):
Ah! Thanks

PR #380, opened 2024-05-13
Fix default progress bar for SignedURL 

kyungeonchoi:
The default progress bar is set to `overall_progress` when `Delivery` is set to `SignedURLs`. Fix default to show progress bar for each sample.


BenGalewsky (approved):
Ok

PR #384, opened 2024-05-17
Check for FORCE_AWKWARD1 to point dependency at awkward v1

holzman:
Coffea 0.7.x needs awkward 1, so it's useful to be able to install serviceX without bringing along conflicting
versions.


ponyisi:
As a general question, why do we depend on awkward in the ServiceX client at all? Client-side data type conversion?

BenGalewsky:
@gordonwatts this is a proposed change to the 2.0 frontend... what do you think?

holzman:
Alternatively, Gordon proposed this:

https://github.com/ssl-hep/ServiceX_frontend/issues/390

ponyisi:
Do we intend any further v2 client changes? Should we just close this now?

PR #387, opened 2024-05-26
Properly handle transform cancellation. Fix tests

ponyisi:
Handle cases where the transformation status is canceled.

Update tests to latest models.


BenGalewsky (changes requested):
You picked up commits from a different PR here. Please rebase and clean up

ponyisi:
Will do. Actually after discussion with @gordonwatts I need to tweak some contracts so I will Draft-ify it for the moment.

ponyisi:
24 hour countdown for comments here :-) The key point here is that cancellation/fatal errors will now raise an exception; we need to decide how far up this propagates, e.g. right now partial success will give a partially-complete output (whereas apparently the 2.0 client would raise an exception).

servicex/query.py
```diff
@@ -63,6 +63,10 @@
 logger = logging.getLogger(__name__)
 
 
+class ServiceXException(Exception):
+    """ Something happened while trying to carry out a ServiceX request """
+
```
gordonwatts, line 68: Didn't this already exist - in `utils.py` - which I can't find now... Perhaps that got lost in the branch.

servicex/query.py
```diff
@@ -354,10 +371,28 @@ async def transform_status_listener(
...
+                elif self.current_status.status == Status.canceled:
+                    logger.warning(
+                        f"Request {titlestr}canceled: "
+                        f"{self.current_status.files_completed}/{self.current_status.files} "
+                        f"files completed"
+                    )
+                    err_str = f"Request {titlestr}was canceled"
+                    if self.current_status.log_url is not None:
+                        err_str += f"\nLogfiles at {self.current_status.log_url}"
+                    raise ServiceXException(err_str)
```
gordonwatts, line 390: The code that generates the message here looks identical to above - and complex. Should this be refactored?

servicex/query.py
```diff
@@ -456,7 +491,8 @@ async def get_signed_url(
             # are guaranteed to be in the bucket. Also, if we are just downloading or
             # signing urls for a previous transform then we know it is complete as well
             if cached_record or (
-                self.current_status and self.current_status.status == Status.complete
+                self.current_status
+                and self.current_status.status in (Status.complete, Status.canceled, Status.fatal)
```
gordonwatts, line 495: I wonder if we shouldn't have a constant that is set `sx_status_done = (Status.complete, Status.canceled, Satus.fatal)` and then we can do things like `self.current_status.status in sx_status_done` or similar - since this gets repeated several times.

gordonwatts (approved):
This looks great! Thank you so much for working on this!!

servicex/query.py
ponyisi: Yup, it got lost in the branch ... I can maybe do a second PR to restore parts of it?

servicex/query.py
ponyisi: Probably a good idea - if we introduce new status codes (e.g. "lookup failed") this could get unwieldy writing it out.

servicex/query.py
ponyisi: I feel it differs enough between the two invocations that refactoring would be rather ugly. If we for some reason wind up using the pattern a third time I would probably reconsider.

BenGalewsky (approved):
For faster turn around on reviews, please don't combine unrelated changes into a single PR or at least make sure the commits make it clear what is going on.

BenGalewsky:
@ponyisi can you fix the conflicts?

PR #391, opened 2024-06-01
Eliminate Pydantic warnings in tests

ponyisi:
Update deprecated APIs to get rid of Pydantic warnings which were polluting test outputs. Change some Mock calls from spec= to return_value= as per its use elsewhere.


BenGalewsky (approved):
Thanks for keeping the campground clean!

PR #394, opened 2024-06-06
Readthedocs deployement .yaml and requirements.txt files

ketan96-m:
Added the .readthedocs.yaml and requirements.txt file for RTD


BenGalewsky (approved):
It would have been useful to have a little more context on the PR. What does this change do?

Issue #382, opened 2024-05-16
3.0 client progress bar issues


ponyisi:
A couple of small complaints about the progress bars in the 3.0 client:
* At least from the command line, they are red until the completion of a request, when they turn green. My general UI expectation is that a red in-progress bar means there has been a failure, but this is not the case here. I would suggest (if possible) that the progress bars are green while the requests are in flight, unless there is in fact a failure.
* Occasionally the download progress bar will not go to full completion even though the downloads are in fact complete. Should ensure that the final update occurs.

PR #397, opened 2024-06-12
Progress bar changes color on completion and failure

ketan96-m:
### Issue
https://github.com/ssl-hep/ServiceX_frontend/issues/382

### Description
The current progress bar is green through out the process.
Introduced a muted green color for normal process and added a brighter green color when transform bar and download bar finishes all files.
If the transformation fails prematurely, then the bar changes the color to red.

Normal progress:
image

Completed progress:
image

Failed Progress:
image

### Solution:
Created a custom `TransformStatusProgress` class which inherits the `Progress` from `rich.progress`.
`get_renderables()` function is overloaded to update the `BarColumn()` with the `red` color in case of failed files.

Added an if condition inside the `query.py`'s `transform_status_listener` that checks that the transform is complete and if there are any failed files and then call the update method for the progress bar.

The update method takes a new keyword argument to indicate if transformation is complete or not.

Added subsequent test cases to test this new `get_renderables()` function and also edited other test cases to with the new `TransformStatusProgress`.


servicex/expandable_progress.py
```diff
@@ -183,3 +184,16 @@ def advance(self, task_id, task_type):
...
             self.progress.advance(task_id=task_id)
+
+
+class TranformStatusProgress(Progress):
+    def get_renderables(self):
+        for task in self.tasks:
+            if task.fields.get("bar") == "failure":
+                self.columns = [
+                    TextColumn("[progress.description]{task.description}"),
+                    BarColumn(complete_style="rgb(255,0,0)"),
```
BenGalewsky, line 195: Why is this here and different from `complete_style="rgb(114,156,31)` above?

servicex/expandable_progress.py
```diff
@@ -74,9 +74,10 @@ def __init__(self,
         self.progress_counts = {}
         if display_progress:
             if self.overall_progress or not provided_progress:
-                self.progress = Progress(
+                self.progress = TranformStatusProgress(
                     TextColumn("[progress.description]{task.description}"),
-                    BarColumn(),
+                    BarColumn(complete_style="rgb(114,156,31)",
```
ketan96-m, line 79: This is the muted green color when the bar is progressing.
image

servicex/expandable_progress.py
ketan96-m: And this red color is for when the bar is progressing but some files have failed.

ponyisi:
Superseded by #408

Issue #395, opened 2024-06-07
Temporary(?) solution for MinIO object store access


kyungeonchoi:
## Problem
- The current approach (list objects in a bucket) to download files or urls from object-store is broken for a bucket more than 1k files ([ref - S3 API Limits](https://min.io/docs/minio/container/operations/checklists/thresholds.html))
- Need a reliable (temporary? but immediate) solution before we have an ultimate solution which requires refactoring of backend and frontend

## Approach
- Explore possible avenues within the same protocol (HTTPS) to have a temporary solution until we have a refactored approach

ponyisi:
One possibility is to switch away from the minio client to something that has a slightly higher level of abstraction: for example, the `boto3` library will handle the pagination in the library.

kyungeonchoi:
@ponyisi - Thanks for your comments!

`boto3` looks cool but it seems like it's a wrapper around `Minio` and uses the same functions. Since our current limitation is coming from the `list_objects()` call, we'd like to explore other options.

I don't know whether partial files (or being written files) in a bucket show up by a `list_objects()` call. We can check this from tests. It might be also useful to check number of files downloaded from MinIO and processed from input dataset (or file list).

ponyisi:
I'm certain boto3 isn't a wrapper around minio (boto3 is the official Amazon library for S3, minio is an independent implementation).

The core point is that there is a very well defined process to list all the files in a bucket, even with minio: you ask for an object list, if you get a thousand objects you ask for the next thousand (using start_after in minio), and keep going until you don't get any more objects. There's no limitation that you can only get 1000 files from a bucket, you just need to make a loop. The value of boto3 is that it would have the code already to do this, so we wouldn't have to write the function ourselves. But, again, there is no fundamental limitation here, just that in writing this code we forgot that this is how minio works.

ketan96-m:
Not all files are uploaded to Minio bucket.
Due to lack of python api for listing al the objects for the minio. I used the CLI `mc` command to manually connect to a bucket when I submitted the transform to check the number of files uploaded. I found that not all files are uploaded to the bucket.

Steps taken:
1. pip install minio
2. https://min.io/docs/minio/linux/reference/minio-mc.html - downloaded the mc tool
3. Ran the bigger_uproot file
```
from servicex import ServiceXSpec, General, Sample
from servicex.func_adl.func_adl_dataset import FuncADLQuery
from servicex.servicex_client import deliver
import logging

[...]

PR #398, opened 2024-06-14
Minio bucket issue when using async_minio

ketan96-m:
Temporary(?) solution for MinIO object store access #395

Open PR for minio bucket issue.
Changed the `minio_adapter.py` to replace the `async_minio` with `minio`

TODO: unit test

### DIFF
--- a/servicex/minio_adapter.py
+++ b/servicex/minio_adapter.py
@@ -30,7 +30,7 @@ from hashlib import sha1
 from pathlib import Path
 from typing import List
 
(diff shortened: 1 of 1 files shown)


ponyisi:
I think we can close this, given the fix in miniopy_async?

kyungeonchoi:
> I think we can close this, given the fix in miniopy_async?

Yes. The fixed version of miniopy_async is also merged into `3.0_develop`

PR #399, opened 2024-06-21
Fix status code enum

ponyisi:
Fix accidental lowercase in the enum definition

### DIFF
--- a/servicex/models.py
+++ b/servicex/models.py
@@ -61,7 +61,7 @@ class Status(str, Enum):
     canceled = ("Canceled",)
     submitted = ("Submitted",)
     looking = ("Lookup",)
-    pending = "Pending lookup"
+    pending = "Pending Lookup"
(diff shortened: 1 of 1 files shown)


kyungeonchoi:
@ponyisi - Does this need to be matched with [the `TransformStatus` class in ServiceX](https://github.com/ssl-hep/ServiceX/blob/58985586f0d06683c4df6e80a3483c274aff3522/servicex_app/servicex/models.py#L148)?

ponyisi:
> @ponyisi - Does this need to be matched with [the `TransformStatus` class in ServiceX](https://github.com/ssl-hep/ServiceX/blob/58985586f0d06683c4df6e80a3483c274aff3522/servicex_app/servicex/models.py#L148)?

They need to at least map the same strings (which they do now, even if the names of the enum variables are slightly different).

PR #401, opened 2024-06-24
Remove databinder directory

kyungeonchoi:
- Migrate DataBinder features under `databinder` directory into the main client codes
- Deprecate `Sample.Function`. Now Python function query is handled by `Sample.Query` as other query languages
- YAML interface for UprootRaw query
- Optional `General.OutputDirectory` writes a yaml containing fileset
- Update examples


ponyisi:
Hi @kyungeonchoi - this is doing a lot of things at once, some of which are conflicting with changes I'm doing (e.g. importing the query classes into the global servicex namespace but changing their names in the import). Could you split into a) the parts that bring the OutputDirectory functionality to the main deliver, b) the examples updates, c) the class name changes?

examples/obsolete/bigger_uproot.py
ponyisi, line 1: Are all these examples actually obsolete? Should they not be deleted then?

servicex/servicex_client.py
```diff
@@ -27,10 +27,12 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 import logging
 from typing import Optional, List, TypeVar, Any, Type, Mapping, Union
+from ccorp.ruamel.yaml.include import YAML
```
ponyisi, line 30: I would move this import into the function that the parser

examples/obsolete/bigger_uproot.py
kyungeonchoi: They are not :)  I was planning to add more examples to replace some of those that are not obsolete (like multi sample and anchor). I agree this is a bit large PR. Let me split this into three pieces as you suggested. But I want all of them merged this week.

servicex/servicex_client.py
```diff
@@ -44,11 +46,42 @@
 T = TypeVar("T")
 logger = logging.getLogger(__name__)
 
+yaml = YAML()
```
ponyisi, line 49: As above, would be better if it's not a global

servicex/servicex_client.py
```diff
@@ -44,11 +46,42 @@
...
     if isinstance(config, Mapping):
+        logger.debug("Config from dictionary")
         config = ServiceXSpec(**config)
+    elif isinstance(config, ServiceXSpec):
+        logger.debug("Config from ServiceXSpec")
+    elif isinstance(config, str) or isinstance(config, Path):
+        logger.debug("Config from file")
+
+        if isinstance(config, str):
+            file_path = Path(config)
```
ponyisi, line 62: Make sure to add appropriate tests

servicex/servicex_client.py
```diff
@@ -82,40 +117,61 @@ def get_codegen(_sample: Sample, _general: General):
                 if sample.Tree:
                     query = query.set_tree(sample.Tree)
                 sample.Query = query
-
+                logger.debug(f"final qastle_query: {sample.Query.generate_selection_string()}")
+            elif isinstance(sample.Query, PythonQuery):
```
ponyisi, line 121: What is the motivation to add this branch? (I thought the interpretation of PythonQuery worked fine already)

servicex/uproot_raw/uproot_raw.py
```diff
@@ -62,3 +64,13 @@ def generate_selection_string(self):
         else:
             final_query = self.query
         return json.dumps([json.loads(_.model_dump_json()) for _ in final_query])
+
+    @classmethod
```
ponyisi, line 68: What does this do? In particular it's not mandatory that every query is a TreeSubQuery?

kyungeonchoi: This allows YAML interface like other queries with yaml tag. You're correct as it assumes it's a `TreeSubQuery`. I added minimal things for now.

kyungeonchoi: Add support for `CopyHistogramSubQuery` by [3c7f25c](https://github.com/ssl-hep/ServiceX_frontend/pull/401/commits/3c7f25c0cb6bf8638127a40020db444b3b81e8d5)

kyungeonchoi:
@ponyisi - is it okay with you to keep the new examples in this PR? And keeping original examples as it is (not under obsolete directory). I think it's useful to bundle them together with the changes to keep track of the working and newly added UX.

servicex/servicex_client.py
kyungeonchoi: Yes! There are couple of places :)

ponyisi:
Let's go for it...

PR #402, opened 2024-06-26
General.ServiceX as an optional field

kyungeonchoi:
- `General.ServiceX` as an optional field to further minimize boilerplate
- Now `General` block can be omitted from `ServiceXSpec` and only `Sample` block is required
- Default `General.ServiceX` value is taken from the `default_endpoint` if present in the `servicex.yaml`. Otherwise the first endpoint is taken. (Assuming normal users have one endpoint)

### DIFF
--- a/servicex/databinder_models.py
+++ b/servicex/databinder_models.py
@@ -96,7 +96,7 @@ class General(BaseModel):
         LocalCache = "LocalCache"
         SignedURLs = "SignedURLs"
 
-    ServiceX: str = Field(..., alias="ServiceX")
+    ServiceX: str = Field(default=None)
     Codegen: Optional[str] = None
     OutputFormat: ResultFormat = (
         Field(default=ResultFormat.root, pattern="^(parquet|root-file)$")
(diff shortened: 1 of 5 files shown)


ponyisi:
Coming back to an earlier discussion - should this just be an argument to `deliver`?

kyungeonchoi:
> Coming back to an earlier discussion - should this just be an argument to `deliver`?

I thought that's for experts - who have multiple endpoints. I expect normal users download `servicex.yaml` file from the web and just use that without knowing the name of ServiceX endpoint. 

I am happy to add an optional argument in the `deliver` function. From 
```
def deliver(config: Union[ServiceXSpec, Mapping[str, Any]], config_path: Optional[str] = None):
```
to 
```
def deliver(config: Union[ServiceXSpec, Mapping[str, Any]], config_path: Optional[str] = None, servicex_name: Optional[str]):
```
to pick up the endpoint from `servicex_name`

ponyisi:
Right, the idea was that if the endpoint is an argument to `deliver` then you get rid of General.ServiceX completely.

ponyisi (approved):

PR #403, opened 2024-06-27
Logging Kibana URL for transformers with failures

ketan96-m:
For transformers which have failed, the user can quickly go to the kibana link created from the terminal.
There are 3 steps:
1. Transformer completes with some failure - display the kibana link with all the level ('INFO' and 'ERROR')
2. Request cancelled - display the kibana link with all the level('INFO' and 'ERROR')
3. Fatal issue - display the kibana link with level('ERROR')

Used the `create_kibana_link_parameters` method from `transformers.py`

To display all the levels, added an IF condition in the this method include or not include log level


kyungeonchoi (approved):
Thanks for the nice improvement! It looks great!

Screenshot 2024-06-28 at 9 19 58 AM

ponyisi:
Dumb question: will the link work on "dumb" terminals?

kyungeonchoi:
@ponyisi - it's a fair question. And the answer is no :( 
There seems to a way to make a clickable link in terminals.

ponyisi:
So I guess the question is, whether to have a fallback for terminals?

ketan96-m:
We can give the option to try the CLI option for each failed transform request.

PR #406, opened 2024-06-28
Add dataset keyword

ponyisi:
Add dataset keyword for samples, for non-Rucio/non-file list datasets


kyungeonchoi (approved):
- Add support to set dataset (`RucioDatasetIdentifier`,`FileListDataset`, `CERNOpenDataDatasetIdentifier`) via yaml tag like Query
- Add support for CERNOpenData
- Different user interface between YAML and ServiceXSpec. E.g. `Dataset: !Rucio` vs. `Dataset: RucioDatasetIdentifier` or `RucioDID`

PR #407, opened 2024-06-28
Don't add failed results to cache

ponyisi:
A miniature variant of #318 : if there are any failures, don't store in the cache at all (as opposed to noting the failure in the cache). Also doesn't include the locking features of #318.


BenGalewsky (approved):
ok - this looks like a welcome improvement.

PR #410, opened 2024-07-04
3.0_refactor-queries

ponyisi:
No longer directly inherit from Query

### DIFF
--- a/examples/Uproot_FuncADL_Dict.py
+++ b/examples/Uproot_FuncADL_Dict.py
@@ -18,5 +18,5 @@ def uproot_funcadl_dict():
     return deliver(spec)
 
 
-if __name__ == "main":
+if __name__ == "__main__":
     uproot_funcadl_dict()
--- a/examples/Uproot_PythonFunction_Dict.py
+++ b/examples/Uproot_PythonFunction_Dict.py
@@ -24,5 +24,5 @@ def uproot_pythonfunction_dict():
     return deliver(spec)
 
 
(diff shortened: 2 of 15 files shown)


ponyisi:
A "decrease" in coverage from this PR is inevitable, since it reduces the number of lines of code in areas which were already covered, while not reducing the number of lines of uncovered code.

BenGalewsky:
> No longer directly inherit from Query

This is such a far-reaching PR. I think it needs a bit more explanation about how to read this. What are we now inheriting from?

ponyisi:
Hi @BenGalewsky - this brings the `FuncADL` and `PythonFunction` query classes into the same framework as the `UprootRaw` query - basically, all the subclasses do now is to provide the query string to send to the server, and the favored codegen. This is done by subclassing the very simple abstract class `QueryStringGenerator`.

The issue is that `Query` has a lot of logic to it, and needs a lot of initialization parameters; if you subclass from it you need to provide those parameters to the subclass, which then just forward it to the superclass initializer, and this is rather fragile (and in any case the subclasses never have a reason to alter any of these). The refactoring fully separates the logic of "how do we produce the query string?" and "how do we execute the query?".

ponyisi:
@BenGalewsky @gordonwatts @kyungeonchoi any additional tests we want to do before deploying this? Everything seems to work on the testing1 instance right now.

kyungeonchoi (approved):
This is amazing! Thanks! Let me wait until later today and merge!

PR #419, opened 2024-07-17
Fix for function indentation in Python transformer

ponyisi:
If source code was indented, it would previously fail with IndentationError. Fix by removing common indentation at the start of lines.


kyungeonchoi (approved):
This works great! Thanks! Let me merge then

Issue #367, opened 2024-05-01
Client 3.0 Exception - ReadTimeOut


kyungeonchoi:
Client library 3.0 gives the `ReadTimeOut` exception intermittently. It seems to happen when a backend or the machine where ServiceX deployed is overloaded with something (not clear the exact condition of ServiceXApp pod). 

 The client I'm testing is the `3.0_develop_uprootraw` branch (fe6a938e75d4c85d6d0f2e0178969a81fd2ab727) but it should behave the same for `3.0_develop` branch.

The following is the last ServiceX client call before `httpx`. I guess the `ServiceXApp` was in a state where it cannot accept any http call. One can play with the timeout setting of http connections but it cannot be an ultimate solution.

```
File /opt/miniconda3/envs/idap/lib/python3.9/site-packages/servicex/servicex_adapter.py:105, in ServiceXAdapter.submit_transform(self, transform_request)
    103 async with httpx.AsyncClient() as client:
    104     headers = await self._get_authorization(client)
--> 105     r = await client.post(url=f"{self.url}/servicex/transformation",
    106                           headers=headers,
    107                           json=transform_request.dict(by_alias=True,
    108                                                       exclude_none=True))
    109     if r.status_code == 401:
    110         raise AuthorizationError(
    111             f"Not authorized to access serviceX at {self.url}")
```

<details>
  <summary>Full stack of error</summary>

```
  ---------------------------------------------------------------------------
ReadTimeout                               Traceback (most recent call last)
File /opt/miniconda3/envs/idap/lib/python3.9/site-packages/httpx/_transports/default.py:69, in map_httpcore_exceptions()
     68 try:
---> 69     yield
     70 except Exception as exc:

File /opt/miniconda3/envs/idap/lib/python3.9/site-packages/httpx/_transports/default.py:373, in AsyncHTTPTransport.handle_async_request(self, request)
    372 with map_httpcore_exceptions():
--> 373     resp = await self._pool.handle_async_request(req)
    375 assert isinstance(resp.stream, typing.AsyncIterable)

File /opt/miniconda3/envs/idap/lib/python3.9/site-packages/httpcore/_async/connection_pool.py:216, in AsyncConnectionPool.handle_async_request(self, request)
    215     await self._close_connections(closing)
--> 216     raise exc from None
    218 # Return the response. Note that in this case we still have to manage
    219 # the point at which the response is closed.

[...]

PR #426, opened 2024-08-22
Robust HTTP connection

kyungeonchoi:
- This PR should resolve issues on timeouts communicating with ServiceX App
- The main change is the HTTP client library: `httpx` to `aiohttp_retry`
- Performance is better with `aiohttp` and various retry mechanism with `aiohttp_retry` package: [more detail](https://oxylabs.io/blog/httpx-vs-requests-vs-aiohttp)
- Implement retry with exponential backoff
- Relevant issues
   - https://github.com/iris-hep/idap-200gbps-atlas/issues/142
   - https://github.com/ssl-hep/ServiceX_frontend/issues/367


ketan96-m:
In light of using the new Aiohttp module for making async get and post requests, the typical httpx.Response was not working.
Since we were dealing with double context managers e.g
```
        async with RetryClient() as client:
            async with client.post(url, headers=headers, json=None) as r:
```

Found a way to patch the get request
Decorate the test functions with
`@patch("servicex.servicex_adapter.RetryClient.get")`

Mock response and status
```
get.return_value.__aenter__.return_value.json.return_value = transform_status_response
get.return_value.__aenter__.return_value.status = 200
```

Resources:
https://stackoverflow.com/questions/28850070/python-mocking-a-context-manager
https://blog.sneawo.com/blog/2019/05/22/mock-aiohttp-request-in-unittests/


Optimizations:
```
get.return_value.__aenter__.return_value.json.return_value = transform_status_response
get.return_value.__aenter__.return_value.status = 200
```
Any help in compressing the 2 lines into one so that we can pass the response(with json and status code) directly to the get.return_value
Typical httpx.Response() doesn't work

BenGalewsky (approved):
Thanks, this looks much more robust now. Nice testing!

PR #427, opened 2024-08-26
Truncate sample name to 128 characters

kyungeonchoi:
- Truncate Sample name to 128 characters if length is longer than 128 characters
- Print warning for truncated sample
- This resolves https://github.com/iris-hep/idap-200gbps-atlas/issues/140

### DIFF
--- a/servicex/databinder_models.py
+++ b/servicex/databinder_models.py
@@ -32,15 +32,18 @@ from pydantic import (
     Field,
     model_validator,
 )
+import logging
 
 from servicex.dataset_identifier import (DataSetIdentifier, RucioDatasetIdentifier,
                                          FileListDataset)
 from servicex.query_core import QueryStringGenerator
 from servicex.models import ResultFormat
 
+logger = logging.getLogger(__name__)
+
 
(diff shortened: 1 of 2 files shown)


servicex/databinder_models.py
```diff
@@ -79,6 +82,18 @@ def validate_did_xor_file(cls, values):
...
 
+    @model_validator(mode="before")
+    @classmethod
+    def truncate_long_sample_name(cls, values):
+        """
+        Truncate sample name to 128 characters if exceed
+        Print warning message
+        """
+        if len(values["Name"]) > 128:
+            print(logger.warning(f"Truncating Sample name to 128 characters for {values['Name']}"))
```
BenGalewsky, line 93: I don't think there should be a print here. `logger.warning` should output to stdout

BenGalewsky (changes requested):

servicex/databinder_models.py
kyungeonchoi: I don't understand why it's there :)

BenGalewsky (approved):
Thanks

Issue #362, opened 2024-04-25
Change naming convention on 3.0 client library


kyungeonchoi:
### Goal
Straightforward and intuitive names for newcomers (especially w/o backend knowledge). 

### Names for discussion

1. `Codegen`
   - We might want to choose what we want to expose to normal users between `codegen` and `transformer`. It can be quite confusing since we often mention different types of transformers
1. Codegen name `uproot`
   - There are `python`, `uproot-raw` and `uproot` for uproot transformer. `uproot` codegen should be something like `uproot-funcadl`. And maybe `uproot-python` for consistency
1. `SignedURLs`
   - More self-explanatory name?
1. `root-file` as `OutputFormat`
   - `-file` only exists for root format
   - It might be good to change to `root-tree` as we will add `root-rntuple`(?) in the future
1. else?

BenGalewsky:
We've agreed that we won't make back-end changes for these, so the unfortunate `root-file` value will have to be translated in the client

kyungeonchoi:
1-2. Stays same
3. URLs
4. `parquet`, `root-ttree` and `root-rntuple`

PR #435, opened 2024-09-04
Change names following #362

ponyisi:
* SignedURLs -> URLs
* Users specify "root-ttree" instead of "root-file" and enum names are changed to "root_ttree".  This involves splitting the types `ResultFormat` and `OutputFormatEnum` : the latter is now used in the `ServiceXSpec` and is translated to a `ResultFormat` when creating the actual request, so that the string sent to the server is still `root-file`. 

Important Note: 
The `ResultFormat` is still what you get back from the server, though, so the cache will continue to handle `root-file` and not `root-ttree`. If at some point we change this string it'll have the side-effect of invalidating caches.

### DIFF
--- a/servicex/databinder_models.py
+++ b/servicex/databinder_models.py
@@ -117,14 +117,8 @@ class General(BaseModel):
         URLs = "URLs"
 
     Codegen: Optional[str] = None
-    OutputFormat: OutputFormatEnum = (
-        Field(default=OutputFormatEnum.root_ttree, pattern="^(parquet|root-ttree)$")
-    )  # NOQA F722
-
-    Delivery: DeliveryEnum = Field(
-        default=DeliveryEnum.LocalCache, pattern="^(LocalCache|URLs)$"
-    )  # NOQA F722
-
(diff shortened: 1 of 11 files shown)


servicex/databinder_models.py
```diff
@@ -98,19 +98,27 @@ def truncate_long_sample_name(cls, values):
 class General(BaseModel):
     class OutputFormatEnum(str, Enum):
         parquet = "parquet"
-        root = "root-file"
+        root_ttree = "root-ttree"
+
+        def to_ResultFormat(self) -> ResultFormat:
```
BenGalewsky, line 103: I think this deserves a comment

BenGalewsky (approved):
Just add a comment to why the `to_result_format` is needed and we are good to go...

ponyisi:
Hold off on a merge, I'm trying to fix something...

ponyisi:
OK, we can continue. I had to remove the redundant string regex validation in the model definition, because that broke the enum coercion; it should be OK because the allowed enum values provide the validation.

Issue #432, opened 2024-09-03
Remove analysis package dependencies from frontend


ponyisi:
The ServiceX frontend should not try to provide data transformation logic on its own, e.g. it should not try to handle turning the returned files into awkward/pandas/... . These dependencies could potentially clash with user choices and should be removed.

ponyisi:
This would obsolete #293

PR #436, opened 2024-09-04
Reduce library dependencies (addresses #432)

ponyisi:
Make `pandas` a dependency only of tests (where it gets used to interpret data), remove references to it from the main library, remove the pandas extra.

Also remove the databinder extra because the package it installs (`nest-asyncio`) is not actually invoked.


BenGalewsky (approved):
Hooray for simplicity!

PR #437, opened 2024-09-04
Improve GuardList traceback handling

ponyisi:
Instead of storing the original exception in the GuardList, wrap the traceback in a new exception type for better debugging


kyungeonchoi (approved):

BenGalewsky (approved):
Yes!

Issue #295, opened 2023-07-14
No network access required for fully cached query


gordonwatts:
User should not need to connect to the internet when accessing a fully cached/downloaded query.

* Example: when you create the dataset, it queries for the code-gen backends

kyungeonchoi:
We need a test for this

BenGalewsky:
This should be the case. Let's verify and see if we can close this

ponyisi:
Added a test, verified code passes (#439 )

PR #439, opened 2024-09-04
Avoid minio connection if cache is sufficient

ponyisi:
Address #295 by only verifying that we only list the S3 buckets when we have a cached record if we are trying to get local files when only signed URLs are cached, or vice versa. Fix a number of tests.


BenGalewsky (approved):
yay, glad this works!

PR #457, opened 2024-09-16
Readthedocs with pyproject.toml

ketan96-m:
Configured the .readthedocs.yaml to run with pyproject.toml
The .yaml file takes the following
```
python:
  install:
    - method: pip
      path: .
      extra_requirements:
        - docs
```
which is equivalent to `pip install .[docs]`
https://docs.readthedocs.io/en/stable/config-file/v2.html#python-install


BenGalewsky (approved):
Thanks for simplifying our build environment

PR #458, opened 2024-09-18
Fix minio handling of buckets in rare cases

ponyisi:
Sometimes listing the bucket for an in-progress transformation gives objects that look like directories, not files. These have no file size reported and would raise an exception due to the coercion in the `ResultFile` constructor. Avoid these problematic objects.

### DIFF
--- a/servicex/minio_adapter.py
+++ b/servicex/minio_adapter.py
@@ -77,7 +77,7 @@ class MinioAdapter:
                 size=obj.size,
                 extension=obj.object_name.split(".")[-1],
             )
-            for obj in objects
+            for obj in objects if not obj.is_dir
         ]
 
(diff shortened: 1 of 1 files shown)


gordonwatts (approved):
This looks fine with me. The only thing that comes to mind is if this gets combined with a "transform done" - so a file gets missed from the list of files reported to the client.

However, I know that is a more general problem and this fixes something specific. So lets approve this, and get it in, and then we'll address that bug when we actually see it.

ponyisi:
Yeah - in any case this is strictly better behavior than we had before, because we would have just crashed and lost everything. I'm not 100% what circumstances lead to this but we should in any case validate the full number of expected files in the bucket at the end of transforms.

Issue #450, opened 2024-09-10
Protect client against bad replies when requesting transform status from server


ponyisi:
When testing against the AGC I see frequent errors of the form
```
  File 
"/home/onyisi/servicex/analysis-grand-challenge/analyses/cms-open-data-ttbar/venv/lib64/python3.9/site-package
s/servicex/query_core.py", line 204, in transform_complete
    raise task.exception()
  File 
"/home/onyisi/servicex/analysis-grand-challenge/analyses/cms-open-data-ttbar/venv/lib64/python3.9/site-package
s/servicex/query_core.py", line 363, in transform_status_listener
    await self.retrieve_current_transform_status()
  File 
"/home/onyisi/servicex/analysis-grand-challenge/analyses/cms-open-data-ttbar/venv/lib64/python3.9/site-package
s/servicex/query_core.py", line 433, in retrieve_current_transform_status
    s = await self.servicex.get_transform_status(self.request_id)
  File 
"/home/onyisi/servicex/analysis-grand-challenge/analyses/cms-open-data-ttbar/venv/lib64/python3.9/site-package
s/servicex/servicex_adapter.py", line 142, in get_transform_status
    o = await r.json()
  File 
"/home/onyisi/servicex/analysis-grand-challenge/analyses/cms-open-data-ttbar/venv/lib64/python3.9/site-package
s/aiohttp/client_reqrep.py", line 1194, in json
    await self.read()
  File 
"/home/onyisi/servicex/analysis-grand-challenge/analyses/cms-open-data-ttbar/venv/lib64/python3.9/site-package
s/aiohttp/client_reqrep.py", line 1134, in read
    self._body = await self.content.read()
  File 
"/home/onyisi/servicex/analysis-grand-challenge/analyses/cms-open-data-ttbar/venv/lib64/python3.9/site-package
s/aiohttp/streams.py", line 383, in read
    block = await self.readany()
  File 
"/home/onyisi/servicex/analysis-grand-challenge/analyses/cms-open-data-ttbar/venv/lib64/python3.9/site-package
s/aiohttp/streams.py", line 405, in readany
    await self._wait("readany")
  File 
"/home/onyisi/servicex/analysis-grand-challenge/analyses/cms-open-data-ttbar/venv/lib64/python3.9/site-package
s/aiohttp/streams.py", line 312, in _wait
    await waiter
aiohttp.client_exceptions.ClientPayloadError: Response payload is not completed: <ContentLengthError: 400, 
[...]

PR #469, opened 2024-09-19
Add retries on the `get` request of `get_transform_status` function

kyungeonchoi:
- Add retries (5 times with 3 sec sleep) on the `get` request of `get_transform_status` function
- This can fix the issue #450

### DIFF
--- a/servicex/servicex_adapter.py
+++ b/servicex/servicex_adapter.py
@@ -135,14 +135,12 @@ class ServiceXAdapter:
         retry_options = ExponentialRetry(attempts=5, start_timeout=3)
         async with RetryClient(retry_options=retry_options) as client:
             try:
-                async for attempt in AsyncRetrying(retry=retry_if_not_exception_type
-                                                   (
-                                                       AuthorizationError,
-                                                       ValueError
-                                                   ),
-                                                   stop=stop_after_attempt(3),
-                                                   wait=wait_fixed(3),
-                                                   reraise=True):
+                async for attempt in AsyncRetrying(
+                        retry=(retry_if_not_exception_type(AuthorizationError)
+                               | retry_if_not_exception_type(ValueError)),
+                        stop=stop_after_attempt(3),
+                        wait=wait_fixed(3),
+                        reraise=True):
                     with attempt:
                         async with client.get(url=f"{self.url}/servicex/"
                                               f"transformation/{request_id}",
--- a/servicex/servicex_adapter.py
+++ b/servicex/servicex_adapter.py
@@ -32,7 +32,7 @@ from typing import Optional, Dict, List
 import httpx
 from aiohttp_retry import RetryClient, ExponentialRetry
 from google.auth import jwt
-from tenacity import AsyncRetrying, stop_after_attempt, wait_fixed
+from tenacity import AsyncRetrying, stop_after_attempt, wait_fixed, retry_if_not_exception_type
 
 from servicex.models import TransformRequest, TransformStatus
 
@@ -132,10 +132,15 @@ class ServiceXAdapter:
 
     async def get_transform_status(self, request_id: str) -> TransformStatus:
         headers = await self._get_authorization()
-        retry_options = ExponentialRetry(attempts=5, start_timeout=10)
+        retry_options = ExponentialRetry(attempts=5, start_timeout=3)
         async with RetryClient(retry_options=retry_options) as client:
             try:
-                async for attempt in AsyncRetrying(stop=stop_after_attempt(5),
+                async for attempt in AsyncRetrying(retry=retry_if_not_exception_type
+                                                   (
+                                                       AuthorizationError,
+                                                       ValueError
+                                                   ),
+                                                   stop=stop_after_attempt(3),
                                                    wait=wait_fixed(3),
                                                    reraise=True):
                     with attempt:
--- a/servicex/servicex_adapter.py
+++ b/servicex/servicex_adapter.py
@@ -32,8 +32,7 @@ from typing import Optional, Dict, List
 import httpx
 from aiohttp_retry import RetryClient, ExponentialRetry
 from google.auth import jwt
-from tenacity import AsyncRetrying, stop_after_attempt, wait_fixed
-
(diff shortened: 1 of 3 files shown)


servicex/servicex_adapter.py
```diff
@@ -133,11 +133,20 @@ async def get_transform_status(self, request_id: str) -> TransformStatus:
...
         async with RetryClient(retry_options=retry_options) as client:
-            async with client.get(url=f"{self.url}/servicex/transformation/{request_id}",
-                                  headers=headers) as r:
-                if r.status == 401:
-                    raise AuthorizationError(f"Not authorized to access serviceX at {self.url}")
-                if r.status == 404:
-                    raise ValueError(f"Transform ID {request_id} not found")
-                o = await r.json()
-                return TransformStatus(**o)
+            retries = 5
```
ponyisi, line 136: I hate to be obnoxious and ask you to learn a new package, but perhaps you can use `tenacity` (https://tenacity.readthedocs.io/en/latest/ , which Ben has been using in the backend code for retries)? it simplifies a lot of this logic, something like

```
from tenacity import AsyncRetrying, RetryError, stop_after_attempt, wait_fixed

try:
    async for attempt in AsyncRetrying(stop=stop_after_attempt(5), wait=wait_fixed(3), reraise=True):
        with attempt:
            async with client.get( ... )
except RuntimeError as e:
     ....
````

Comment (ponyisi) [reply servicex/servicex_adapter.py:136]: (I've been asking for tenacity >= 9.0.0)

**Review by ponyisi — CHANGES_REQUESTED** [review]

File: servicex/servicex_adapter.py (hunk 1 above)
Comment (gordonwatts) [reply servicex/servicex_adapter.py:136]: I agree here - please use a package that has done this - there are some really nice ones out there - and they handle all the logic we could imagine way better than us. `tenacity` is pretty nice!!

File: servicex/servicex_adapter.py [hunk 2]
```diff
@@ -133,11 +134,21 @@ async def get_transform_status(self, request_id: str) -> TransformStatus:
...
-            async with client.get(url=f"{self.url}/servicex/transformation/{request_id}",
-                                  headers=headers) as r:
-                if r.status == 401:
-                    raise AuthorizationError(f"Not authorized to access serviceX at {self.url}")
-                if r.status == 404:
-                    raise ValueError(f"Transform ID {request_id} not found")
-                o = await r.json()
-                return TransformStatus(**o)
+            try:
+                async for attempt in AsyncRetrying(stop=stop_after_attempt(5),
```
ketan96-m, line 138: Shouldn't we add the retry for a particular error like `ClientPayloadError`(from the issue). 

AsyncRetrying() has parameter `retry` which can retry if a particular exception has occured using the `retry=retry_if_exception_type(ClientPayloadError)`

servicex/servicex_adapter.py
```diff
@@ -149,6 +148,6 @@ async def get_transform_status(self, request_id: str) -> TransformStatus:
                                 raise ValueError(f"Transform ID {request_id} not found")
                             o = await r.json()
                             return TransformStatus(**o)
-            except RuntimeError as e:
+            except RetryError as e:
```
ketan96-m, line 151: Added a RetryError exception handling instead of RuntimeError since that will capture the last retried error from the AsyncRetrying

servicex/servicex_adapter.py
ponyisi: I would put this as a 3.0.1 improvement instead, I'd rather just brute force this for now and be more subtle later.

servicex/servicex_adapter.py
ponyisi: Not if we use `reraise=True`, no?

ketan96-m: For a short code block for the try catch, RuntimeError makes sense.  If the block was bigger we could have multiple exception handled.
Although I tested this using a small test block, both work fine. However, for now I will keep this RuntimeError exception.
I will revert the change.

gordonwatts (changes requested):
I think there is only one change that needs to happen now - the 10 second minimum for the backoff. I'm worried how long 5 failures will take if you have something like transform not found (which happens when you re-run things). Also, please create an issue for us to restrict the retry exceptions. The blanket means anything else goes wrong and the system will "hang" while it keep re-trying.

tests/test_servicex_adapter.py
```diff
@@ -199,6 +198,13 @@ async def test_get_transform_status_auth_error(get, servicex):
         assert "Transform ID b8c508d0-ccf2-4deb-a1f7-65c839eebabf not found" == str(err.value)
 
 
+@pytest.mark.asyncio
+async def test_get_tranform_status_retry_error(servicex):
+    with pytest.raises(Exception) as err:
+        await servicex.get_transform_status("b8c508d0-ccf2-4deb-a1f7-65c839eebabf")
+        assert "ServiceX WebAPI Error while getting transform status:" in str(err.value)
```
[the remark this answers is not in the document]
ketan96-m: Can it because no Runtime error was raised, we can have an explicit ClientError exception handling.

servicex/servicex_adapter.py
ponyisi: Shouldn't we just alter the retry statement now to exclude the `AuthorizationError` and `ValueError` exceptions with `retry_if_not_exception_type`? Seems easier than waiting.

kyungeonchoi: @gordonwatts - the keyword argument `reraise=True` of `AsyncRetrying()` enables to capture the original error. Thus no need to worry about swallowed errors.

servicex/servicex_adapter.py
```diff
@@ -133,11 +134,21 @@ async def get_transform_status(self, request_id: str) -> TransformStatus:
         headers = await self._get_authorization()
         retry_options = ExponentialRetry(attempts=5, start_timeout=10)
```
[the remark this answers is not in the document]
kyungeonchoi: That's correct - Let me set the timeout to 3 sec and exclude AuthorizationError and ValueError for retry

gordonwatts (approved):
This looks good now. Please add an issue for follow up.

PR #470, opened 2024-09-19
Resolves the progress bar final count not updating

### LINKED ISSUE #464: The marching ants do not deal with evolving total numbers of files
Running on a vary large dataset means that it takes `servicex` a while to determine the complete list of files. Foe example, `data18_13TeV:data18_13TeV.periodAllYear.physics_Main.PhysCont.DAOD_PHYSLITE`. This has some 64K files in it. While the dashboard seems to report things (mostly) correctly:

image

The marching ants seems to take the first number back from SX and report that for the complete run (rather than updating):

image

It can take quite a while to determine what the list of files is - so this means the marching ants are incorrect!

ketan96-m:
Resolves #464
#### Bug:
image

The marching ants seems to take the first number back from SX and report that for the complete run (rather than updating):

image

#### Solution:
There was a bug in the implementation of the update code of progress bar. Fixed with to take the total files from the backend.

Added a test case for advancing the progress bar.


ketan96-m:
@gordonwatts Can you please review

gordonwatts (approved):

PR #478, opened 2024-09-25
Update the download progress bar (both unsigned and dowload) when initial transform request doesn't return all files

ketan96-m:
### Problem
When the user submits many requests, sometimes the transform status returns only partial number of files.
After subsequent polling, the correct number of files are returned. However, the download progress bar wasn't geting updated.
image

### Solution
Inside the `transform_status_listener` added the update logic for download progress bar.


ketan96-m:
@ponyisi ready for review

ponyisi (approved):
OK

PR #481, opened 2024-09-27
Add XRootD dataset identifier class

### LINKED ISSUE #477: Add XRootD dataset lookup as a first-class dataset object
We are missing a class in `servicex.dataset` corresponding to the XRootD DID finder lookup. We should fix that.

ponyisi:
Add a class to encapsulate an XRootD dataset request (resolve #477)

### DIFF
--- a/docs/datasets.rst
+++ b/docs/datasets.rst
@@ -1,0 +1,30 @@
+Specifying input datasets
+========
+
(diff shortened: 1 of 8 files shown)


gordonwatts (approved):
this looks good. But might need better eyes, @ponyisi

PR #482, opened 2024-09-28
Cache Collision when the user ignores the cache and hash is present in the db Fix #375

### LINKED ISSUE #375: Cache Collisions when you ignore cache for a re-run
This is with the current 3.0 alpha 16.

To repro:

1. Run a query
2. Re-run query and tell the system to ignore the cache
3. Re-run the query again, this time without the ignore.

You'll get a cache collision error:

```text
0000.0845 - INFO - root - Using release 22.2.107 for type information.
0000.1185 - WARNING - func_adl.type_based_replacement - Unknown type for name len
0000.8609 - INFO - root - Building ServiceX query
0000.8610 - INFO - root - Using dataset mc20_13TeV.364157.Sherpa_221_NNPDF30NNLO_Wmunu_MAXHTPTV0_70_CFilterBVeto.deriv.DAOD_PHYSLITE.e5340_s3681_r13145_p6026.
0000.8611 - INFO - root - Running on 10 files of dataset.
0000.8615 - INFO - root - Starting ServiceX query
0000.9198 - INFO - servicex.servicex_client - Returning code generators from cache

Traceback (most recent call last):
  File "/home/gwatts/code/iris-hep/idap-200gbps-atlas/servicex/servicex_materialize_branches.py", line 356, in <module>
    main(
  File "/home/gwatts/code/iris-hep/idap-200gbps-atlas/servicex/servicex_materialize_branches.py", line 119, in main
    files = query_servicex(
  File "/home/gwatts/code/iris-hep/idap-200gbps-atlas/servicex/servicex_materialize_branches.py", line 92, in query_servicex
    results = sx.deliver(spec)
  File "/venv/lib/python3.9/site-packages/servicex/servicex_client.py", line 107, in deliver
    results = group.as_signed_urls()
  File "/venv/lib/python3.9/site-packages/make_it_sync/func_wrapper.py", line 63, in wrapped_call
    return _sync_version_of_function(fn, *args, **kwargs)
  File "/venv/lib/python3.9/site-packages/make_it_sync/func_wrapper.py", line 14, in _sync_version_of_function
    return loop.run_until_complete(r)
  File "/usr/AnalysisBaseExternals/25.2.2/InstallArea/x86_64-el9-gcc13-opt/lib/python3.9/asyncio/base_events.py", line 647, in run_until_complete
    return future.result()
  File "/venv/lib/python3.9/site-packages/servicex/dataset_group.py", line 76, in as_signed_urls_async
    return await asyncio.gather(*self.tasks)
  File "/venv/lib/python3.9/site-packages/servicex/query.py", line 521, in as_signed_urls_async
    return await self.submit_and_download(
  File "/venv/lib/python3.9/site-packages/servicex/query.py", line 210, in submit_and_download
    self.cache.get_transform_by_hash(sx_request.compute_hash())
  File "/venv/lib/python3.9/site-packages/servicex/query_cache.py", line 84, in get_transform_by_hash
    raise CacheException("Multiple records found in db for hash")
servicex.query_cache.CacheException: Multiple records found in db for hash
```

ketan96-m:
### Problem
Resolve #375 

```
cached_record = (
            self.cache.get_transform_by_hash(sx_request_hash)   # this raises exception when 2 records are present
            if not self.ignore_cache
            else None
        )
```
When the user submits a transform without `ignore-cache` -> records are inserted as expected
When the user submits same transform with `ignore-cache` -> a NEW record is still created (tiny db has no unique value contraints)
When the user submits same tranform without `ignore-cache` -> 2 records are present hence raises exception

### Fix

When the user submits a query and selects ignore_cache -
Check if we have a previous hash already present and if present delete that record since we are going to run the transform again.

### DIFF
--- a/servicex/query_core.py
+++ b/servicex/query_core.py
@@ -228,11 +228,16 @@ class Query:
                 logger.info(f"Transforms finished with code {self.current_status.status}")
 
         sx_request = self.transform_request
+        sx_request_hash = sx_request.compute_hash()
+
+        # Invalidate the cache if the hash already present but if the user ignores cache
+        if self.ignore_cache and self.cache.contains_hash(sx_request_hash):
+            self.cache.delete_record_by_hash(sx_request_hash)
 
         # Let's see if this is in the cache already, but respect the user's wishes
         # to ignore the cache
         cached_record = (
-            self.cache.get_transform_by_hash(sx_request.compute_hash())
+            self.cache.get_transform_by_hash(sx_request_hash)
             if not self.ignore_cache
             else None
         )
--- a/tests/test_servicex_dataset.py
+++ b/tests/test_servicex_dataset.py
@@ -30,6 +30,7 @@ from typing import List
 from unittest.mock import AsyncMock
 from pathlib import PurePath
 import pytest
+from itertools import cycle
 
 from servicex.configuration import Configuration
 from servicex.dataset_identifier import FileListDataset
@@ -149,6 +150,20 @@ transform_status6 = transform_status.model_copy(
         "files": 2,
     }
 )
+
+
+@pytest.fixture
+def servicex():
+    servicex = AsyncMock()
+    servicex.submit_transform = AsyncMock()
+    servicex.submit_transform.return_value = {"request_id": '123-456-789"'}
+    servicex.get_transform_status = AsyncMock()
+    # servicex.get_transform_status.side_effect = [
+    #     transform_status1,
+    #     transform_status3,
+    # ]
+    yield servicex
+
 
 file1 = ResultFile(filename="file1", size=100, extension="parquet")
 file2 = ResultFile(filename="file2", size=100, extension="parquet")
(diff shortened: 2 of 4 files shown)


ponyisi:
Could you add a test for the specific code path that codecov is complaining about? Then I'll approve

servicex/query_core.py
```diff
@@ -228,11 +228,16 @@ def transform_complete(task: Task):
                 logger.info(f"Transforms finished with code {self.current_status.status}")
 
         sx_request = self.transform_request
+        sx_request_hash = sx_request.compute_hash()
+
+        # Invalidate the cache if the hash already present but if the user ignores cache
+        if self.ignore_cache and self.cache.contains_hash(sx_request_hash):
```
gordonwatts, line 234: Any reason not to just do the check always - `if self.cache.contains_hash(sx_request_hash)` instead? That is - don't protect it by `ignore_cache`?

ketan96-m: If the user submits the same request again, then the cache will already have the record, so it will recompute rather than use the record from the cache. We only want to invalidate if the record is present but the user still wants to recompute (ignore cache).

ketan96-m:
@ponyisi @gordonwatts ready for review
The unit tests were a piece to get it working!

servicex/query_core.py
ketan96-m: Also, I have not the subsequent logic, so I found it best to include this at a top without breaking anything below

tests/test_servicex_dataset.py
```diff
@@ -527,3 +526,82 @@ def test_transform_request():
                               "(lambda (list e) (dict (list 'lep_pt') " \
                               "(list (subscript e 'lep_pt')))))"
         cache.close()
+
+
+@pytest.mark.asyncio
+async def test_use_of_cache_ignore_cache(mocker):
+    """ Do we pick up the cache on the second request for the same transform? """
+    servicex = AsyncMock()
```
BenGalewsky, line 534: Could you pull these mocks out into a [pytest fixture](https://docs.pytest.org/en/stable/explanation/fixtures.html)? It will make this test cleaner and easier to read... also we will hopefully need these for other tests

BenGalewsky (approved):
ok - thanks for fixing this annoying bug

PR #484, opened 2024-09-30
expose OutputFormatEnum instead of ResultFormat to user

### LINKED ISSUE #447: `ResultFormat` should be removed
The `servicex` package still exports `ResultFormat`, which is an enum defined in the `models.py` file. I suspect other things like `ResultDesitination` need to also be removed.

ponyisi:
Resolves #447 by changing the enum that is given to the user to be the one used by Pydantic to validate user input, instead of the one that defines interactions with the backend


gordonwatts (approved):
nice.will this break Api?

ponyisi:
> nice.will this break Api?

Yes. However it would only ever have worked for someone who used parquet (since by definition root was broken) and I don't think it's actually documented anywhere in our user guide (!)

PR #485, opened 2024-09-30
Prevent user passing nfiles=0 when creating a sample

### LINKED ISSUE #467: You can specify 0 files for the number of files to transform
The interface should either crash if you ask for zero files, or it should return an empty transform properly. Currently a hang occurs, and the UX is very confused.

ponyisi:
Resolve #467 by rejecting setting `NFiles=0` (or `num_files=0`) during `Sample` specification


gordonwatts (approved):
Nice!

PR #486, opened 2024-10-02
Multiple samples submitted together updates the database only once #417

### LINKED ISSUE #417: Messed up cache database when multiple samples containing the same file is requested
```
from servicex import ServiceXSpec, General, Sample, deliver

def run_query(input_filename=None):
    import uproot
    #
    return uproot.open({input_filename: "CollectionTree"}).arrays(['EventInfoAux.eventNumber'])

filename = ("root://eosuser.cern.ch//eos/atlas/atlascerngroupdisk/phys-higgs/HSG1/MxAOD/h028/mc16a/Nominal/"
            "mc16a.PowhegPy8_bbH125.MxAODDetailedNoSkim.e6050_s3126_r9364_p4180_h028.root")

spec = ServiceXSpec(
    General=General(
        ServiceX="servicex-uc-af",
        Codegen="python",
        OutputFormat="root-file",
        Delivery="LocalCache",
    ),
    Sample=[
        Sample(
            Name="foo",
            XRootDFiles=[filename],
            Function=run_query
        ),
        Sample(
            Name="bar",
            XRootDFiles=[filename],
            Function=run_query
        )
    ]
)
deliver(spec)
```

This will give
```
{'foo': ['/tmp/servicex_chlcheng/1e39cdcb-8c78-46ac-9eee-9b23961d3c0e/_fb8d56b26c71d7760000ede59ac8cb59ead5b951364_p4180_h028.root'],
 'bar': ['/tmp/servicex_chlcheng/f62efa75-184a-490d-b77a-d77b5cda3db6/_fb8d56b26c71d7760000ede59ac8cb59ead5b951364_p4180_h028.root']}
```

Rerunning the request will give the error
```
File /pscratch/sd/c/chlcheng/local/miniconda/envs/ml-gpu/lib/python3.11/site-packages/servicex/query_cache.py:89, in QueryCache.get_transform_by_hash(self, hash)
     86     return None
     88 if len(records) != 1:
---> 89     raise CacheException("Multiple records found in db for hash")
     90 else:
     91     return TransformedResults(**records[0])

CacheException: Multiple records found in db for hash
```

ketan96-m:
Problem #417 
Fixes #417 

When a sample has many files which have the same details except different sample names, multiple records were created for the same request (same hash). Due to the async nature of the requests, the files are downloaded asynchronously and hash value was also inserted in the db.

Fix
Added a condition in the insert statement of the cache where it only inserts if the hash value is not previously present. The locks previously installed eliminates the race condition.

### DIFF
--- a/servicex/query_cache.py
+++ b/servicex/query_cache.py
@@ -70,7 +70,8 @@ class QueryCache:
 
     def cache_transform(self, record: TransformedResults):
         with self.lock:
-            self.db.insert(json.loads(record.model_dump_json()))
+            if not self.contains_hash(record.hash):
+                self.db.insert(json.loads(record.model_dump_json()))
 
     def update_record(self, record: TransformedResults):
         transforms = Query()
--- a/tests/test_query_cache.py
+++ b/tests/test_query_cache.py
@@ -32,7 +32,7 @@ import pytest
 
 from servicex.configuration import Configuration
 from servicex.models import ResultFormat
-from servicex.query_cache import QueryCache, CacheException
+from servicex.query_cache import QueryCache
 from tinydb import Query
 
 file_uris = ["/tmp/foo1.root", "/tmp/foo2.root"]
@@ -103,15 +103,7 @@ def test_cache_transform(transform_request, completed_status):
             )
         )
 
-        with pytest.raises(CacheException):
-            _ = cache.get_transform_by_hash(transform_request.compute_hash())
-
-        with pytest.raises(CacheException):
-            _ = cache.get_transform_by_request_id(
-                "b8c508d0-ccf2-4deb-a1f7-65c839eebabf"
-            )
-
-        assert len(cache.cached_queries()) == 2
+        assert len(cache.cached_queries()) == 1
 
         cache.delete_record_by_request_id("b8c508d0-ccf2-4deb-a1f7-65c839eebabf")
         assert len(cache.cached_queries()) == 0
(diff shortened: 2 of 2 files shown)


ketan96-m:
@ponyisi @BenGalewsky @gordonwatts ready for review!

servicex/query_cache.py
```diff
@@ -70,7 +70,8 @@ def transformed_results(self, transform: TransformRequest,
 
     def cache_transform(self, record: TransformedResults):
         with self.lock:
-            self.db.insert(json.loads(record.model_dump_json()))
+            if not self.contains_hash(record.hash):
+                self.db.insert(json.loads(record.model_dump_json()))
```
gordonwatts, line 74: Just want to make sure - is there a race condition here? That is- do you need to `lock` the DB to make this query so no one inserts something between line 73 and 74?

gordonwatts (approved):
I'm ok with this as long as we aren't setting ourselves up for a race condition here!

servicex/query_cache.py
ponyisi: I'm confused - this _is_ already inside a lock, no? (`self.lock`)

gordonwatts: Ah - I thougth that was to prevent multi-threaded inside the single app, I was worried we were running multiple servicex app's.

gordonwatts: And this is the front end - not the ackend sign. Right - this looks great.

ponyisi: Also it's a file lock, so it should actually work across multiple processes. (Which actually sets us up for an interesting race condition on its own ... @ketan96-m , does this only kick in once the submission to the backend is complete? Or is it added to the db before submission?)

ketan96-m: It is kicked in after the backend is complete

PR #489, opened 2024-10-03
Expose `Delivery` enum

### LINKED ISSUE #449: `OutputFormatEnum` and `DeliveryEnum` are not exposed to the library user
Currently, you can't easily do something like:

```python
        General=sx.General(
            Codegen=query[1],
            OutputFormat=sx.OutputFormatEnum.root_ttree,  # Fails with error
            Delivery=(DeliveryEnum.LocalCache if download else sx.DeliveryEnum.URLs),  # type: ignore
        ),
```

because it isn't exported from the top level. Rather you must delve quite deep into the object hierarchy. They should be exported at the top level (the way `ResultFormat` and `ResultDesitnation` were previosly.

ponyisi:
Resolve #449 by exposing `DeliveryEnum` as `servicex.Delivery` (`OutputFormat` was already added). Add documentation.


kyungeonchoi (approved):
Looks great! Thanks

PR #491, opened 2024-10-03
docs: Note that ReadTheDocs is the user docs

matthewfeickert:
* Distinguish between the 'user' docs being the default 'stable' view of ReadThe Docs and the 'developer' docs being the 'latest' view, which is built from the latest commit on the default branch.


matthewfeickert:
This is ready for review.

ponyisi (approved):

PR #494, opened 2024-10-05
Resume request when client fails fixes #468

### LINKED ISSUE #468: Frontend crashes while working transform is in progress are not saved
The following sequence:

1. Your code submits a transform.
2. While waiting for it to finish, you have a crash or disconnect
3. Re-run, and a new transform is fired off, even if the old transform is still running and doing just fine.

Expected behavior:

* Same transform should always be used
* Running transform that hasn't failed should be reused.

ketan96-m:
Resolves #468
### Problem
When the client submits the request but encounters client failure, there is no way of resuming the transforms from the last point. The client is forced to rerun the request which triggers the backend again.

### Fix
This PR fixes the broken client problem. The client will be able to resume the request from the last point. Given the client submits the same request again (important since the request hash is calculated and compared with any pending requests)
Future scope:
- Provide a CLI option to rerun all the jobs which are stale due to client failures
- Download the files which are already processed by the backend by providing a request id (CLI)

### Approach
Created a new instance of tinydb whose sole purpose is to keep track of the transform request submissions. 
I had a discussion with @ponyisi whether we should have the cache use additional columns and flags. The cache currently holds the TransformationResults AFTER the request is processed. We could technically have additional columns, but it would mean changing previous cache specific functions.
I propose to create a new instance of tinydb which stores the submitted transform request ( I have named it as `queue`). This will allow to add transform request submission related functions to be separate from the cache features. Thus keeping the existing code intact and only adding the necessary steps in between.
Also this will allow us to add the future scope (CLI) option without much modification.

Open to design inputs

### Details of this PR
- Created a separat instance of tinydb called `queue` which records each new transform_request
- Incase of multiple transform request with same hash only one record is maintained and only one submission to backend is created saving redundant requests
- Incase of failure on the client side, the queue resumes when you run the same request again. However, if you submit is with 'ignore cache` parameter, it will rerun the request to the backend.

### DIFF
--- a/servicex/query_cache.py
+++ b/servicex/query_cache.py
@@ -126,6 +126,11 @@ class QueryCache:
         else:
             return TransformRequest(**records[0])
 
+    def queue_delete_record(self, record: TransformRequest):
+        hash_value = record.compute_hash()
+        with self.lock:
+            self.queue.remove(where('hash') == hash_value)
+
     def cache_transform(self, record: TransformedResults):
         with self.lock:
             if not self.contains_hash(record.hash):
--- a/servicex/query_cache.py
+++ b/servicex/query_cache.py
@@ -215,13 +215,3 @@ class QueryCache:
         with self.lock:
             self.db.remove(where('backend') == backend)
 
-    async def get_request_id(self, request: TransformRequest):
-        transform = Query()
-        with self.lock:
-            hash = request.compute_hash()
-            record = self.db.search(transform.hash == hash)
-
-        if not record:
-            return None
-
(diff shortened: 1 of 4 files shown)


ketan96-m:
@ponyisi ready for review

BenGalewsky:
I'm sorry, @ketan96-m - I don't agree with @ponyisi on this... adding a new database just to avoid updating existing code is just a recipe for technical debt. Please add a new boolean column for `completed` which indicates if this request has reached an end one way or another. That way you can add entries as soon as they are submitted.

ponyisi:
Superseded by #497

PR #497, opened 2024-10-11
Single Database instance -- Resume request when client fails fixes #468 

### LINKED ISSUE #468: Frontend crashes while working transform is in progress are not saved

ketan96-m:
Resolves https://github.com/ssl-hep/ServiceX_frontend/issues/468

Problem
When the client submits the request but encounters client failure, there is no way of resuming the transforms from the last point. The client is forced to rerun the request which triggers the backend again.

Fix
Single Database instance -- Added flag
This PR fixes the broken client problem. The client will be able to resume the request from the last point. Given the client submits the same request again (important since the request hash is calculated and compared with any pending requests)

### DIFF
--- a/servicex/query_cache.py
+++ b/servicex/query_cache.py
@@ -84,21 +84,24 @@ class QueryCache:
         """
         transforms = Query()
         with self.lock:
-            records = self.db.search((transforms.hash == hash) & (transforms.files.exists()))
+            records = self.db.search((transforms.hash == hash)
+                                     & ~(transforms.status == 'SUBMITTED'))
         return len(records) > 0
 
-    def get_transform_request_status(self, hash_value: str) -> Optional[str]:
+    def is_transform_request_submitted(self, hash_value: str) -> bool:
         """
         Get the status of a transform request
         """
         transform = Query()
         with self.lock:
-            records = self.db.search((transform.hash == hash_value)
-                                     & (transform.status.exists()))
+            records = self.db.search((transform.hash == hash_value))
 
         if not records:
-            return None
-        return records[0]['status']
+            return False
+
+        if "status" in records[0] and records[0]["status"] == 'SUBMITTED':
+            return True
+        return False
 
     def get_transform_request_id(self, hash_value: str) -> Optional[str]:
         """
@@ -137,7 +140,7 @@ class QueryCache:
         transforms = Query()
         with self.lock:
             records = records = self.db.search((transforms.hash == hash)
-                                               & (transforms.files.exists()))
+                                               & ~(transforms.status == 'SUBMITTED'))
 
         if not records:
             return None
--- a/servicex/query_cache.py
+++ b/servicex/query_cache.py
@@ -79,12 +79,18 @@ class QueryCache:
             self.db.update(json.loads(record.model_dump_json()), transforms.hash == record.hash)
 
     def contains_hash(self, hash: str) -> bool:
+        """
+        Check if the cache has completed records for a hash
+        """
         transforms = Query()
         with self.lock:
             records = self.db.search((transforms.hash == hash) & (transforms.files.exists()))
         return len(records) > 0
 
     def get_transform_request_status(self, hash_value: str) -> Optional[str]:
+        """
+        Get the status of a transform request
+        """
         transform = Query()
         with self.lock:
             records = self.db.search((transform.hash == hash_value)
@@ -95,6 +101,9 @@ class QueryCache:
         return records[0]['status']
 
     def get_transform_request_id(self, hash_value: str) -> Optional[str]:
+        """
+        Return the request id of cached record
+        """
         transform = Query()
 
         with self.lock:
@@ -105,17 +114,26 @@ class QueryCache:
         return records[0]["request_id"]
 
     def update_transform_status(self, hash_value: str, status: str) -> None:
+        """
+        Update the cached record status
+        """
         transform = Query()
         with self.lock:
             self.db.upsert({"hash": hash_value, "status": status}, transform.hash == hash_value)
 
     def update_transform_request_id(self, hash_value: str, request_id: str) -> None:
+        """
+        Update the cached record request id
+        """
         transform = Query()
         with self.lock:
             self.db.upsert({"hash": hash_value, "request_id": request_id},
                            transform.hash == hash_value)
 
     def get_transform_by_hash(self, hash: str) -> Optional[TransformedResults]:
+        """
+        Returns completed transformations by hash
+        """
         transforms = Query()
         with self.lock:
             records = records = self.db.search((transforms.hash == hash)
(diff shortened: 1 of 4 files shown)


servicex/query_core.py
```diff
@@ -285,7 +285,25 @@ def transform_complete(task: Task):
         )
 
         if not cached_record:
-            self.request_id = await self.servicex.submit_transform(sx_request)
+            ###
+            # if the transform request is pending
```
BenGalewsky, line 289: Please don't check in commented out code. If we need this code it is in the commit history

ketan96-m:
@BenGalewsky, @ponyisi, @gordonwatts  ready for review!

kyungeonchoi:
Just had a chat with @ketan96-m that we need to check how download behaves when resume (or rerun `deliver()`) ServiceX request. We need to make sure it recognizes the `data_dir` of original request and resume download from where it lost connection.

ponyisi (commented):
General comments following discussion -
* only add to cache when request ID is obtained
* do not worry about edge case where two requests of the same hash are made in the same Spec (we could instead raise an error if this happens, but it is almost certainly a user mistake)
* ensure that if the transform fails, we wipe the in-progress cache record
* things are bit copy-and-paste in a few places, reduce this

Otherwise the overall structure looks fine to me

ponyisi:
> Just had a chat with @ketan96-m that we need to check how download behaves when resume (or rerun `deliver()`) ServiceX request. We need to make sure it recognizes the `data_dir` of original request and resume download from where it lost connection.

Hi @kyungeonchoi if the user changes the spec of the request and therefore the target file area, I think we should just go ahead and deliver there. I believe with this PR we will just redownload all files from the incomplete requests (regardless of what has already been downloaded) which is a bit brute force but can be improved in the future - I don't think we need to include "avoid re-downloading" yet, since that really will require us to do an integrity check on the files that we have already.

gordonwatts (commented):
This looks great. I think my questions are just because I don't understand something deeper going on. Please take a look and then decide!

gordonwatts:
It seems like the `transform_complete` has become a very large method with many many lines. It might be time to add it to the list of things that need to be refactored.

servicex/query_cache.py
```diff
@@ -81,13 +82,56 @@ def update_record(self, record: TransformedResults):
     def contains_hash(self, hash: str) -> bool:
         transforms = Query()
         with self.lock:
-            records = self.db.search(transforms.hash == hash)
+            records = self.db.search((transforms.hash == hash) & (transforms.status.exists())
+                                     & (transforms.status == 'COMPLETE'))
```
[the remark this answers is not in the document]
ketan96-m: Interesting, I changed the condition so that rather than the status flag, it checks if the files exists or not, that way it is backward compatible

servicex/query_cache.py
```diff
@@ -69,25 +69,75 @@ def transformed_results(self, transform: TransformRequest,
...
             self.db.update(json.loads(record.model_dump_json()), transforms.hash == record.hash)
 
     def contains_hash(self, hash: str) -> bool:
+        """
+        Check if the cache has completed records for a hash
+        """
         transforms = Query()
         with self.lock:
-            records = self.db.search(transforms.hash == hash)
+            records = self.db.search((transforms.hash == hash) & (transforms.files.exists()))
```
ponyisi, line 87: I *strongly* prefer this test is on `transforms.status != 'SUBMITTED'`. As it is, we are relying on the fact that certain information is not in a record to infer the type of the record, when we actually do have the relevant field!

servicex/query_cache.py
```diff
@@ -69,25 +69,75 @@ def transformed_results(self, transform: TransformRequest,
...
         return len(records) > 0
 
+    def get_transform_request_status(self, hash_value: str) -> Optional[str]:
+        """
+        Get the status of a transform request
+        """
+        transform = Query()
+        with self.lock:
+            records = self.db.search((transform.hash == hash_value)
+                                     & (transform.status.exists()))
```
ponyisi, line 97: Mild preference to do this check *after* the query, i.e. look up the request first (because having the request missing should be an error!) and then check if the field is there.

It's also true that you basically never use this function except to check if the result is `SUBMITTED`. Perhaps better to make the function `get_transform_request_submitted` and return a bool?

servicex/query_cache.py
```diff
@@ -69,25 +69,75 @@ def transformed_results(self, transform: TransformRequest,
...
+
     def get_transform_by_hash(self, hash: str) -> Optional[TransformedResults]:
+        """
+        Returns completed transformations by hash
+        """
         transforms = Query()
         with self.lock:
-            records = self.db.search(transforms.hash == hash)
+            records = records = self.db.search((transforms.hash == hash)
+                                               & (transforms.files.exists()))
```
ponyisi, line 140: See comment from line ~ 87

ponyisi (commented):
In the interest of time let's approve and

servicex/query_cache.py
ponyisi: I guess maybe it's `~(transforms.status == 'SUBMITTED')` ...

servicex/query_cache.py
ketan96-m: I have changed the function to a bool. However, the request missing shouldn't raise an error since the the request is still not submitted in the case and it should return False rather than a error.

tests/test_dataset.py
```diff
@@ -277,6 +278,8 @@ async def test_submit_and_download_cache_miss_overall_progress(python_dataset, c
         python_dataset.servicex.submit_transform = AsyncMock()
         python_dataset.download_files = AsyncMock()
         python_dataset.download_files.return_value = []
+        python_dataset.cache.update_transform_request_id = Mock()
+        # python_dataset.cache.update_transform_status = Mock()
```
ponyisi, line 282: Remove commented code?

ponyisi (approved):
See one line to update, otherwise I approve

servicex/query_cache.py
```diff
@@ -81,13 +82,56 @@ def update_record(self, record: TransformedResults):
...
+            with self.lock:
+                records = self.db.search(transform.hash == hash_value)
+
+            if not records:
+                return None
+
+            if len(records) != 1:
+                raise CacheException("Multiple records found in db for hash")
+            else:
+                if 'request_id' in records[0]:
```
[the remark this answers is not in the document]
ponyisi: This is a "key present" query not a substring query

servicex/query_cache.py
ponyisi: Right, I see. Somewhat unfortunate naming of the function I suppose. It would be good to document this contract (it returns False if the transformation is not in the cache at all) in the docstring, since you rely on this behavior.

tests/test_dataset.py
ketan96-m: Done!

PR #508, opened 2024-11-05
Add some documentation on errors

### LINKED ISSUE #496: Understand/Document current contract around incomplete/failed transforms
We need documentation around how the front end behaves when a transform is incomplete or fails.

* When not all files are transformed (e.g. 1000 files in original dataset, 998 successful, 2 fail). What happens?
* When the submission fails (e.g. codegen throws)

If during documentation something is clearly off, then either fix or create new issue!

ponyisi:
Add some documentation for the user documenting our contract for error handling in the client. We may want to revisit some of these before the 3.0.1 release. Resolves #496


ponyisi:
Codecov is a false positive

kyungeonchoi (approved):
Looks great! Thanks!

PR #510, opened 2024-11-05
Improve handling of error codes from server

### LINKED ISSUE #476: Better document YAML definition and include features
Better explain the anchor/alias and `!include` functionality of the YAML file parsing in the frontend documentation.

### LINKED ISSUE #502: HTTP error handling in client often implicitly expects response to be JSON
There are cases like https://github.com/ssl-hep/ServiceX_frontend/blob/7105fd358f2373cb4f5e07e0fede30ead256769b/servicex/servicex_adapter.py#L127 where we assume we can call `r.json()` when the server response may not be JSON (this is essentially a certainty for 500 errors for example). The retries will hide this at the moment because they're not picky about the exception type, but if there is a real server error that needs to get propagated to users, this behavior will obfuscate it. 

Plan: go through code and fix places where we assume a JSON response in abnormal conditions.

ponyisi:
Resolve #502. Do not assume that error responses from the backend are JSON except for specific HTTP codes.

### DIFF
--- a/servicex/query_core.py
+++ b/servicex/query_core.py
@@ -197,7 +197,8 @@ class Query:
             expandable_progress.refresh()
             if task.exception():
                 logger.error(
-                    "ServiceX Exception", exc_info=task.exception()
+                    f"ServiceX Exception for request ID {self.request_id} ({self.title})\"",
+                    exc_info=task.exception()
                 )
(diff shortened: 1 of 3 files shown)


BenGalewsky (approved):
Ahhh, nice to see this annoying bug fixed

PR #511, opened 2024-11-05
Bump release version to 3.0.1-alpha.1

ponyisi:
Prepare for 3.0.1 release with an alpha

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "servicex"
-version = "3.0.0-alpha.100"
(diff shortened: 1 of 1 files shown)


BenGalewsky:
@ponyisi we don't need to change the version via commit. This is [set during the release process](https://github.com/ssl-hep/ServiceX_frontend/blob/226d3dca163adaf80034cf18f0c999e9130e4785/.github/workflows/pypi.yaml#L43)

ponyisi:
I had issues installing from a git repo, because the existing installed version 3.0.0 was "newer" than the git repo one. So I'd like to keep this at least ahead of the released version even if it doesn't actually affect the published package.

PR #512, opened 2024-11-05
Require tree name to be set for `FuncADLQuery_Uproot` queries

### LINKED ISSUE #452: Warn users if they attempt a FuncADL_Uproot submission with no tree defined
Users need to give a `.FromTree` selection when using FuncADL_Uproot. Error if this is missing so they don't have to puzzle through some bizarre error message on the transformer side.

ponyisi:
Resolves #452. Throw an exception if `FuncADLQuery_Uproot` is used without specifying a tree name.


BenGalewsky (approved):
Cool!

PR #513, opened 2024-11-05
Require namespace in Rucio DIDs

### LINKED ISSUE #31: Require namespace in dataset
Bomb if a namespace isn't present in a rucio ds name.

Add this fact to the docs

ponyisi:
Resolves #31. `RucioDatasetIdentifier` will throw an exception if a namespace isn't specified.


BenGalewsky (approved):
Yes!

Issue #210, opened 2021-11-02
Misleading error message with missing `servicex.yaml` 


alexander-held:
In the absence of a `servicex.yaml` file defining endpoints, I see the following error message:

> ServiceXException: (ServiceXException(...), 'Unable to find name/type uproot in api_endpoints in servicex.yaml configuration file. Saw only names (default) and types (xaod)')

This sounds as if a `servicex.yaml` file was found, but that my desired endpoint was not included in there. Instead, it would be useful if the error message mentioned that no file was found.

ponyisi:
Fixed in 3.0 client

PR #514, opened 2024-11-05
Add a test to make sure that we raise an exception if there is no servicex config file

ponyisi:
Test to make sure that #210 (which is already fixed) stays fixed


BenGalewsky (approved):
Ok!

PR #516, opened 2024-11-06
New argument `fail_if_incomplete` to raise an exception if any files fail

ponyisi:
Change `deliver()` to fail for a sample (by default) if any files fail. Can be disabled by passing `fail_if_incomplete=False`.


kyungeonchoi (approved):
This is good.

This PR changes a returned value by `deliver()`. It returns an exception rather than a list of only successful files when there is failed file.

kyungeonchoi:
It would be nice to propagate this change into the documentation PR #508

PR #520, opened 2024-11-15
Do a preflight authentication call

### LINKED ISSUE #506: Do a pre-flight authentication check
As is, validation of the user token is only done when we start trying to submit jobs, and failures are kind of hidden (they're treated the same way as a transformation failure and only visible when you look at the return value of `deliver()`). Instead, we should do an authentication check before starting transform request submission, and immediately raise `AuthorizationError` if a problem is found.

ponyisi:
Resolves #506. Get authentication before actually doing submission, so we can raise an exception early and not hide it in the `GuardList` in the return value of `deliver`.


ponyisi:
I think the codecov issue is a false positive.

kyungeonchoi (approved):
Great to have this fix! Thanks!

PR #521, opened 2024-11-15
Delete a Transform 

BenGalewsky:
# Problem
Users can't delete unused transform requests to free up disk space

# Approach
This PR goes along with [ServiceX PR 915](https://github.com/ssl-hep/ServiceX/pull/915) to provide a command line tool to delete transforms on the server


ponyisi:
So, to bring up the language issue again here ... nothing here removes the local copies of transform results, yes? Is "delete" the correct keyword then? Not maybe "archive" or something similar?

I assume we have an equivalent command to clear a transform from the local cache - do we want to connect the two commands in some way?

BenGalewsky:
Good call on the deleting the cache... pushed a new commit to do so

ponyisi:
I'd like @gordonwatts and @kyungeonchoi to comment on the contract here:
* deleting a transform on the client side removes files locally and on the server
* however the server will have its own archiving mechanism in play which will potentially remove the files remotely 
so I think this is perhaps best thought of (from the user side) as a local file cleanup mechanism, and not as a way to do server file management? (which I guess users shouldn't be worried about anyway)

BenGalewsky:
To be honest, this feature was mostly a stepping stone to the automated lifecycle management, but if you are a polite user and you ran a transform and realized you made a mistake and will never use the results, you can issue this command and it will delete the files locally _and_ off of the server.

If you don't do this, then yes eventually the lifecycle ops will delete this for you

ponyisi:
This probably is best brainstormed at the retreat, but - the issue as I see it is that a very typical pattern is users just running many different variants of a transform to develop it, so at the end they have a lot of old versions, but they haven't kept track of the obsolete development ones, and they aren't likely to go hunt down the old transform IDs one by one.

If I had to envision a more user-friendly scenario, it would be
* client knows the most recent iteration of transforms with a certain name
* there is a command that allows us to purge the previous versions while keeping the latest ones
* the transform deletion command also takes the sample name as an option instead of the transform ID

ponyisi (approved):

PR #522, opened 2024-11-15
Remove view source code button from documentation

kyungeonchoi:
I don't see a reason why user needs to see source code of documentation readthedoc pages. It's a bit annoying place by default, so I would like to remove it 😄 

Current
Screenshot 2024-11-15 at 3 22 06 PM

New
Screenshot 2024-11-15 at 3 24 53 PM


ponyisi (approved):
Fair enough

Issue #351, opened 2024-03-22
Add pytest tests for CLI API


matthewfeickert:
In PR #349, @matthewfeickert forgot to add in the CLI API entrypoint and that had to be patched in PR #350. :grimacing:

It would be good to add tests of the CLI API behavior to ensure that things are working as expected. This can be done using [`pytest-console-scripts`](https://github.com/kvas-it/pytest-console-scripts) and `click.testing.CliRunner`. c.f. [`pyhf`'s tests](https://github.com/scikit-hep/pyhf/blob/428c9e1ae4927f612b4d98da96923e0208339c6c/tests/test_scripts.py) for examples of using this.

BenGalewsky:
Completed via #536

PR #523, opened 2024-11-15
Introduce example of app testing with patches

ponyisi:
Use pytest-console-scripts to allow us to test the `servicex` CLI with patches to backend code. This is not at all intended to be comprehensive, but to establish a framework. Starts to address #351


BenGalewsky (approved):
ok - looks like a good start

Issue #465, opened 2024-09-19
Docs: Update About Page


matthewfeickert:
In the `3.0_develop` branch the [About Page](https://servicex-frontend.readthedocs.io/en/latest/about.html) in the docs is very outdated. It doesn't tell people who don't already know about the IRIS-HEP Slack how to get help

https://github.com/ssl-hep/ServiceX_frontend/blob/21aea65baf20e40c5e6f29249f64637f3423dee9/docs/about.rst?plain=1#L5-L8

and the demos/tutorials listed

https://github.com/ssl-hep/ServiceX_frontend/blob/21aea65baf20e40c5e6f29249f64637f3423dee9/docs/about.rst?plain=1#L26-L32

are 4 years old and so almost certainly not correct anymore. There are also newer demos.

ponyisi:
Any chance you could take this one, @gordonwatts ?

kyungeonchoi:
Resolved by https://github.com/ssl-hep/ServiceX_frontend/pull/526

PR #526, opened 2024-11-19
Update About page of documentation

kyungeonchoi:
- A table of presentations on the client only (or major topic is client) for users who want to know more about the client
- Response to the issue by @matthewfeickert - https://github.com/ssl-hep/ServiceX_frontend/issues/465

### DIFF
--- a/docs/about.rst
+++ b/docs/about.rst
@@ -1,32 +1,53 @@
-About Page
+Outreach
 ===========
 
+This page lists presentations and posters on the ServiceX client (frontend) library.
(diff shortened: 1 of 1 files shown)


matthewfeickert:
Thanks @kyungeonchoi. You can probably also use the `Resolves #465` here as I think you'll do a great job of addressing things here.

ponyisi (approved):
This is good for now. We should provide a feedback channel (which used to be the #servicex channel on Slack) but perhaps we decide on this during the workshop this week.

Issue #528, opened 2024-11-24
`servicex dataset delete` CLI command reports "Dataset N not found"


ponyisi:
Running `servicex dataset delete N` seems to always print "Dataset N not found" even though the delete succeeds. I suspect this is because the CLI uses the return value of `ServiceXClient.delete_dataset` to determine what it prints
https://github.com/ssl-hep/ServiceX_frontend/blob/c4c8562687927f5dd27089cd033145875672a33e/servicex/app/datasets.py#L122
and following the call chain, it's the return value of `ServiceXAdapter.delete_dataset`
https://github.com/ssl-hep/ServiceX_frontend/blob/c4c8562687927f5dd27089cd033145875672a33e/servicex/servicex_adapter.py#L174
which never actually has a path that returns anything other than None.

BenGalewsky:
Resolve via #529

PR #529, opened 2024-11-25
Check for return value on dataset delete

BenGalewsky:
# Problem
The delete command in the dataset cli tests for a True return value from the ServiceXAdapter. Unfortunately, the `delete_dataset` method doesn't return anything.

# Approach
Add a check for the `stale` property in the value returned from the server and return that from the method


ponyisi (approved):
Looks good

ponyisi:
Resolves #528, correct?

PR #532, opened 2024-12-05
Avoid race condition between status monitoring and download tasks

ponyisi:
If a transform task completed while the download polling was going on, we might miss some outstanding files to download. Do not terminate the polling loop now until all files are seen in the object store.


gordonwatts (approved):
The fact we have a hang, with no info for the user, makes me uncomfortable. However, lets get this out.

ponyisi:
I propose to resolve @gordonwatts 's concern in 3.0.2 (#533).

PR #535, opened 2024-12-12
Fix merge mishap with uri option

BenGalewsky:


gordonwatts (approved):
Ah - this will also fix the url/backend both specified problem? :-) Looks great!

PR #537, opened 2024-12-19
Allow user specified Adaptor and MiniIO Interfaces

gordonwatts:
Smaller changes to allow user to specify an Adaptor and MiniIO adaptor in the `deliver` call to affect different backend behavior.


gordonwatts:
Going another direction it looks like!

PR #538, opened 2025-01-01
Register backends on the fly

gordonwatts:
This code was designed to enable a local version of SX to run, in process, to help wtih debugging and development.

* Adds extra object to the `Endpoint` to track a special adaptor.
* Ability to register a new end point on the fly, which then gets appended to the list of read end points.
* `ServiceXClient` properly creates a specialized client when requested., and one for the `minio` `for_transform`.
* Add `ServiceXAdaptor` and `MinioAdaptor` as a protocal.

## Notes:

* `pydantic` can't seem to deal with duck typing Protocals, so `object` is getting used.
* It might be ok to remove the `protocol` totally - especially if the `pydantic` problem can't be solved. Or make them sub-classes of a ABC.


gordonwatts:
Decided to not go this route - and move towards dealing with local as a standalone package.

PR #549, opened 2025-01-17
docs: Update supported Python version lower bound

matthewfeickert:
* Update required Python version lower bound to Python 3.9 in the docs.
* Update supported Python metadata.
* Update the dev container's Python to 3.13, the latest supported Python.

PR #552, opened 2025-01-18
Fix devcontainer python image base

ponyisi:
Previous PR #549 unintentionally changed the devcontainer `Dockerfile` to pull from a non-existent tag (0-3.13). I have changed it to 1-3.13 .


matthewfeickert (approved):
Sorry about that. I falsely assumed that the non-default tags would be consistent across Python versions.

PR #553, opened 2025-01-18
Don't poll S3 unless we have reason to think there are new files available

ponyisi:
Keep the minio polling interval but do nothing if the current transform status does not indicate more files than the last time we looked in S3.


BenGalewsky (approved):
Thanks for knocking this out. I looked to see if we could have a unit test for this condition, but it would require some changes to the code to be able to update the number of files in between polls..
I'll leave that for another day...

PR #554, opened 2025-01-20
Update ci_production.yaml to print target of test

ponyisi:
Make it clearer what is going on in test

The endpoint will be printed in the test. As it is a secret, to get it to output it is encoded via `xxd -ps`;  it can be de-encoded by `xxd -r -ps`.


BenGalewsky (approved):
Ah, nice one

PR #556, opened 2025-01-21
Give users options to condense or eliminate progress bars

ponyisi:
* Fix issues in ExpandableProgress that broke summary bar
* Pass necessary arguments down from deliver()
* Introduce enum class that lets user choose default display (all bars), compact display (only the total progress), or no progress bars at all

### DIFF
--- a/servicex/expandable_progress.py
+++ b/servicex/expandable_progress.py
@@ -147,7 +147,12 @@ class ExpandableProgress:
             task_id = self.progress.add_task(
                 param, start=start, total=total, visible=False
             )
-            new_task = ProgressCounts(param, task_id, start=start, total=total)
+            new_task = ProgressCounts(
+                "Transform" if param.endswith("Transform") else param,
+                task_id,
+                start=start,
+                total=total,
+            )
             self.progress_counts[task_id] = new_task
             return task_id
         if self.display_progress and not self.overall_progress:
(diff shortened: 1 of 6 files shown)


servicex/servicex_client.py
```diff
@@ -45,12 +45,19 @@
...
+from enum import Enum
 import traceback
 
 T = TypeVar("T")
 logger = logging.getLogger(__name__)
 
 
+class ProgressBarFormat(str, Enum):
+    default = "default"
+    compact = "compact"
```
gordonwatts, line 57: Add `expanded`, and have `default` be the same as `expanded` - that makes this stable against future changes for people that want it.

servicex/expandable_progress.py
```diff
@@ -156,13 +161,15 @@ def add_task(self, param, start, total):
     def update(self, task_id, task_type, total=None, completed=None, **fields):
 
         if self.display_progress and self.overall_progress:
+            if task_type.endswith("Transform"):
```
gordonwatts, line 164: I didn't check the code out - but you always seem to `and` `display_progress` and `overall_progress` - which I don't totally get - because you set one or both in the `deliver` function.  With this wouldn't there only be one behavior?

gordonwatts (commented):
This looks fine to me - one suggestion and one question (because I didn't check out the code and test it in a notebook).

servicex/expandable_progress.py
ponyisi: Not far below (line 202) we have an `and` of `display_progress` and `not overall_progress` :-) One could in principle wrap the whole thing in an `if self.display_progress:` block but I didn't really want to play much with the existing code beyond what absolutely needed changing.

servicex/servicex_client.py
ponyisi: OK

gordonwatts (approved):
Fantastic!

ponyisi:
I will add user documentation in a new PR - this is actually somewhat involved but should be done properly.

PR #557, opened 2025-01-21
Remove unused func_adl code

ponyisi:
Remove some code that is not hit by any `Query` code path


BenGalewsky (approved):
I was never brave enough to change any of this code. Thank you for your service!

PR #558, opened 2025-01-23
`deliver()` documentation + some small fixes

ponyisi:
* Add documentation for deliver()
* Change function order in the autodoc
* Add `__version__` to `servicex`
* Show contents of `servicex` module before subpackages
* Add `expanded` to `ProgressBarFormat`
* Bump default git version to 3.1.1-alpha1


servicex/servicex_client.py
```diff
@@ -214,15 +221,38 @@ def _output_handler(
 
 
 def deliver(
-    config: Union[ServiceXSpec, Mapping[str, Any], str, Path],
+    spec: Union[ServiceXSpec, Mapping[str, Any], str, Path],
```
gordonwatts, line 224: Is this a breaking change. ;-) But I think `spec` is way better!

gordonwatts (approved):

servicex/servicex_client.py
ponyisi: It is a breaking change if anyone is passing `config=` as a keyword argument, yeah... none of our examples do this though. Should definitely warn people about it when the release is made.

PR #559, opened 2025-01-23
Improve API documentation

ponyisi:
Improve the presentation of the `servicex` API documentation:
* move the top-level module into its own page
* move the internals to their own file
* keep top-level `servicex.app` and `servicex.func_adl` indices
* use `enum-tools` to make enums look nicer


matthewfeickert:
@ponyisi sorry that I didn't get to this today. I'll try to get a local review done on Friday before late afternoon.

BenGalewsky:
Is there any way to make the `DeliveryEnum` type in the `General` constructor argument documentation clickable?
i.e.
```
Delivery – (DeliveryEnum) Specifies the delivery method for the output files.
```

ponyisi:
@BenGalewsky I've tried a new variant where the "constructor" isn't documented at all for the Pydantic classes (this is the default, we only introduced that block because of trying to make command-line docstrings work nicely).

gordonwatts (approved):
Very hard to see what changes are actually happening by looking at the source code, but nothing stood out.

Issue #525, opened 2024-11-16
Fix issues identified by bugbear report


ponyisi:
I ran flake8-bugbear (https://github.com/PyCQA/flake8-bugbear) on the frontend code and got:

```
servicex/app/cache.py:76:25: B008 Do not perform function calls in argument defaults.  The call is performed only once at function definition time. All calls to your function will reuse the result of that definition-time function call.  If this is intended, assign the function call to a module-level variable and use that variable as a default value.
servicex/app/cache.py:88:32: B008 Do not perform function calls in argument defaults.  The call is performed only once at function definition time. All calls to your function will reuse the result of that definition-time function call.  If this is intended, assign the function call to a module-level variable and use that variable as a default value.
servicex/app/datasets.py:47:37: B008 Do not perform function calls in argument defaults.  The call is performed only once at function definition time. All calls to your function will reuse the result of that definition-time function call.  If this is intended, assign the function call to a module-level variable and use that variable as a default value.
servicex/app/datasets.py:52:40: B008 Do not perform function calls in argument defaults.  The call is performed only once at function definition time. All calls to your function will reuse the result of that definition-time function call.  If this is intended, assign the function call to a module-level variable and use that variable as a default value.
servicex/app/datasets.py:95:27: B008 Do not perform function calls in argument defaults.  The call is performed only once at function definition time. All calls to your function will reuse the result of that definition-time function call.  If this is intended, assign the function call to a module-level variable and use that variable as a default value.
servicex/app/datasets.py:119:27: B008 Do not perform function calls in argument defaults.  The call is performed only once at function definition time. All calls to your function will reuse the result of that definition-time function call.  If this is intended, assign the function call to a module-level variable and use that variable as a default value.
servicex/app/main.py:56:31: B008 Do not perform function calls in argument defaults.  The call is performed only once at function definition time. All calls to your function will reuse the result of that definition-time function call.  If this is intended, assign the function call to a module-level variable and use that variable as a default value.
servicex/app/transforms.py:60:32: B008 Do not perform function calls in argument defaults.  The call is performed only once at function definition time. All calls to your function will reuse the result of that definition-time function call.  If this is intended, assign the function call to a module-level variable and use that variable as a default value.
servicex/app/transforms.py:87:25: B008 Do not perform function calls in argument defaults.  The call is performed only once at function definition time. All calls to your function will reuse the result of that definition-time function call.  If this is intended, assign the function call to a module-level variable and use that variable as a default value.
[...]

PR #560, opened 2025-01-24
Clean up bugbear issues (address #525)

ponyisi:
Address #525 - mostly clean up function calls in argument defaults in CLI app. Bugbear has opinions about "bad code smell" and in this case the warnings are pretty harmless (these are functions that only get called once) but still better to have a clean report.


gordonwatts (approved):
These changes look fine with me.

PR #565, opened 2025-02-26
Remove obsolete fields in transform requests

ponyisi:
Neither `tree-name` nor `image` should be specified in ServiceX requests. The former does nothing anyway and the latter is a security risk. In any case the frontend v3 API doesn't provide any way for users to set them.


kyungeonchoi (approved):
Looks good! Thanks!

PR #566, opened 2025-03-05
WIP: Add TopCP transformer query

kyungeonchoi:
- TopCP transformer is a new transformer in ServiceX that runs the [ATLAS TopCPToolkit framework](https://topcptoolkit.docs.cern.ch)
- This PR adds features to run TopCP transformer
- The ServiceX backend version 1.X.X is required to submit TopCP query

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -104,6 +104,7 @@ FuncADL_ATLASxAOD = "servicex.func_adl.func_adl_dataset:FuncADLQuery_ATLASxAOD"
 FuncADL_CMS = "servicex.func_adl.func_adl_dataset:FuncADLQuery_CMS"
 PythonFunction = "servicex.python_dataset:PythonFunction"
 UprootRaw = "servicex.uproot_raw.uproot_raw:UprootRawQuery"
+TopCP = "servicex.topcp.topcp:TopCPQuery"
 
 [project.entry-points.'servicex.dataset']
 Rucio = "servicex.dataset_identifier:RucioDatasetIdentifier"
--- a/servicex/topcp/topcp.py
+++ b/servicex/topcp/topcp.py
@@ -1,0 +1,129 @@
+# Copyright (c) 2025, IRIS-HEP
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# * Redistributions of source code must retain the above copyright notice, this
+#   list of conditions and the following disclaimer.
+#
+# * Redistributions in binary form must reproduce the above copyright notice,
+#   this list of conditions and the following disclaimer in the documentation
+#   and/or other materials provided with the distribution.
+#
+# * Neither the name of the copyright holder nor the names of its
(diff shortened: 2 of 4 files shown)


servicex/topcp/topcp.py
```diff
@@ -0,0 +1,137 @@
...
+            with open(self.parton_yaml, "r") as parton_file:
+                partonYaml = yaml.safe_load(parton_file)
+
+        particleYaml = None
+        if self.particle_yaml:
+            with open(self.particle_yaml, "r") as particle_file:
+                particleYaml = yaml.safe_load(particle_file)
+
+        query = {
+            "RecoYAML": recoYaml,
```
ponyisi, line 117: Q on this -
* maybe it would be easier just to call it "reco", "parton", "particle" ?
* I guess we should assume that if the user provides a YAML file they wish to run that particular type of analysis? so by default if someone provides a particle YAML then RunParticle should be True?

kyungeonchoi: > * maybe it would be easier just to call it "reco", "parton", "particle" ?

I agree. Let me also change camel case field names to snake case.

> * I guess we should assume that if the user provides a YAML file they wish to run that particular type of analysis? so by default if someone provides a particle YAML then RunParticle should be True?

I'm following the default values of TopCPToolkit except `no_systematics`. I'm open to change the defaults but I guess it's common to have all `reco.yaml`, `parton.yaml` and `particle.yaml` in the same directory and mostly run `reco.yaml` only.

ponyisi: Oh? How does TopCP determine which one(s) to run?

It just seems to me that there is unlikely to be any situation where someone provides a particle-level YAML but doesn't expect the particle analysis to be run by default... ?

pyproject.toml
```diff
@@ -124,3 +125,6 @@ packages = ["servicex"]
 
 [tool.coverage.run]
 dynamic_context = "test_function"
+
+[tool.pytest.ini_options]
+asyncio_mode = "auto"
```
ponyisi, line 130: Not that I'm complaining, but did you find  you needed this?

servicex/topcp/topcp.py
kyungeonchoi: That's fair - TopCP doesn't take each yaml input file. A user provides a directory which contains all yaml files instead. I think your proposal makes more sense as ServiceX query. Let me drop options like `run_parton`.

pyproject.toml
kyungeonchoi: Pytest failed for `tests/test_servicex_adapter.py` with the following error:
[`E           ConnectionResetError: [Errno 104] Connection reset by peer`](https://github.com/ssl-hep/ServiceX_frontend/actions/runs/13773510333/job/38517405365#step:6:111)
Those two lines automatically add the _asyncio_ marker to all async test functions. I have to admit that I don't know whether the error was just an intermittent error and it can be fixed by this.

ponyisi (approved):

ponyisi:
Let's goooo

PR #572, opened 2025-03-21
Switch S3 client to boto3

### LINKED ISSUE #571: Switch to aioboto3 for S3 downloads
The Minio client has essentially no error-checking on downloads and can appear to succeed when it has in fact only partially downloaded a file. These operations are much more carefully implemented in the boto family of libraries. Since the S3 access is encapsulated fully in one class this should not be too painful to do.

ponyisi:
Fixes #571 . `aioboto3` has better built-in error detection and retry support than `minio`.
Merge after #573


BenGalewsky (approved):
This looks great! Thank you

PR #573, opened 2025-03-21
Use semaphore to limit concurrent downloads in non-CLI client

ponyisi:
By default, only run 8 downloads at once in a session (can be overridden with an argument to `deliver`).


BenGalewsky:
Is this PR still valid, given the changes in #571

BenGalewsky:
It certainly needs a good rebase to bring it back in line with the boto3 changes...

PR #574, opened 2025-03-25
Use uv to set up python version in tests

ponyisi:
There is a conflict in setting up Python versions with 3.13t if we use both the setup-python and setup-uv actions. Do only the latter. (This PR does not itself bring in 3.13t, that has conflicts with dependencies, specifically aiohttp https://github.com/aio-libs/aiohttp/discussions/10467 .)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -31,7 +31,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.13t"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
     runs-on: ${{ matrix.platform }}
 
     steps:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -31,7 +31,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, windows-latest, macos-latest]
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -37,18 +37,15 @@ jobs:
...
         uses: astral-sh/setup-uv@v5
+        with:
+          python-version: ${{ matrix.python-version }}
 
       - name: Install dependencies
         run: |
-          uv pip install --system --upgrade '.[test,docs]'
-          uv pip list --system
+          uv pip install --upgrade '.[test,docs]'
+          uv pip list
```
matthewfeickert, line 48: Ah that's interesting. I didn't realize that if you setup Python with `uv` that it knows how to automatically activate the default virtual environment it creates. 

https://github.com/astral-sh/setup-uv/blob/22695119d769bdb6f7032ad67b9bca0ef8c4a174/README.md#python-version

That's good to know, so thanks @ponyisi!

matthewfeickert (approved):

matthewfeickert (commented):
Ah, @ponyisi do you want to also apply these changes to all the other GitHub Action workflows? 

* https://github.com/ssl-hep/ServiceX_frontend/blob/828820277d23713ec157e6aa55a541911c3f2f42/.github/workflows/ci_production.yaml#L30-L36
* https://github.com/ssl-hep/ServiceX_frontend/blob/828820277d23713ec157e6aa55a541911c3f2f42/.github/workflows/docs.yml#L32-L38
* https://github.com/ssl-hep/ServiceX_frontend/blob/828820277d23713ec157e6aa55a541911c3f2f42/.github/workflows/pypi.yaml#L25-L31

?

BenGalewsky (approved):
Looks good

PR #579, opened 2025-04-04
Upgrade local download concurrency control to use aioboto3's built-in handling

ponyisi:
Replace #573 by using the native concurrency handling of the boto3 libraries. Up the default concurrency to 10 (matches the default in the library).

### DIFF
--- a/servicex/minio_adapter.py
+++ b/servicex/minio_adapter.py
@@ -40,7 +40,7 @@ from servicex.models import ResultFile, TransformStatus
 _transferconfig = TransferConfig(max_concurrency=10)
 
 
-def init_download_semaphore(concurrency: int = 10):
+def init_s3_config(concurrency: int = 10):
     "Update the number of concurrent connections"
     global _transferconfig
     _transferconfig = TransferConfig(max_concurrency=concurrency)
--- a/servicex/minio_adapter.py
+++ b/servicex/minio_adapter.py
@@ -30,20 +30,20 @@ from hashlib import sha1
 from pathlib import Path
 from typing import List
 
-import asyncio
 from tenacity import retry, stop_after_attempt, wait_random_exponential
 
 import aioboto3
+from boto3.s3.transfer import TransferConfig
 
 from servicex.models import ResultFile, TransformStatus
 
-_semaphore = asyncio.Semaphore(8)
+_transferconfig = TransferConfig(max_concurrency=10)
 
 
-def init_download_semaphore(concurrency: int = 8):
+def init_download_semaphore(concurrency: int = 10):
     "Update the number of concurrent connections"
-    global _semaphore
-    _semaphore = asyncio.Semaphore(concurrency)
+    global _transferconfig
+    _transferconfig = TransferConfig(max_concurrency=concurrency)
 
 
 def _sanitize_filename(fname: str):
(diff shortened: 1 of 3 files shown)


ponyisi:
Not the world's worst suggestion (named it slightly differently)

> * [nitpick] The function name 'init_download_semaphore' is no longer descriptive since it now configures a TransferConfig rather than initializing a semaphore. Consider renaming it to 'init_transfer_config' for clearer intent.
> 
> ```
> def init_download_semaphore(concurrency: int = 10):
> ```

servicex/minio_adapter.py
```diff
@@ -114,31 +114,29 @@ async def download_file(
...
     @retry(
         stop=stop_after_attempt(3), wait=wait_random_exponential(max=60), reraise=True
     )
     async def get_signed_url(self, object_name: str) -> str:
-        async with _semaphore:
-            async with self.minio.client("s3", endpoint_url=self.endpoint_host) as s3:
-                return await s3.generate_presigned_url(
-                    "get_object", Params={"Bucket": self.bucket, "Key": object_name}
-                )
+        async with self.minio.client("s3", endpoint_url=self.endpoint_host) as s3:
```
gordonwatts, line 136: Will this do ok even when we have a bucket with 50K files in it?

tests/test_minio_adapter.py
```diff
@@ -160,6 +160,25 @@ async def test_download_short_filename_change(minio_adapter, populate_bucket):
     result.unlink()  # it should exist, from above ...
 
 
+@pytest.mark.parametrize("populate_bucket", ["test.txt"], indirect=True)
+@pytest.mark.asyncio
+async def test_download_repeat(minio_adapter, populate_bucket):
+    import asyncio
+
+    print(await minio_adapter.list_bucket())
+    result = await minio_adapter.download_file("test.txt", local_dir="/tmp/foo")
```
gordonwatts, line 169: You parameterize this test with populate_bucket as test.txt, but have it hardcoded here too.

gordonwatts (changes requested):
Minor comments except for oen weird thing in the test where you have a parameterized test function, but then the parameter value hard coded.

servicex/minio_adapter.py
ponyisi: This will get throttled on the server side, as before. We can discuss putting in a separate semaphore for that, but the main motivation for this PR is to avoid problems overwhelming the client.

(boto3 doesn't provide a built-in mechanism for limiting the signed URL rates, which is a bit of an indication that Amazon doesn't really see this as a problem for their services)

tests/test_minio_adapter.py
ponyisi: The parametrization parametrizes the argument to the fixture `populate_bucket`; I'm not aware of a mechanism to make it available to the test function itself (other than doing an additional parametrization with the same argument). In general since we have a fair number of cases where the two arguments are different I don't know that it's super useful to do that, the values should be pretty clear.

servicex/minio_adapter.py
gordonwatts: On to testing, then!

gordonwatts (approved):
looks good, thanks for the answers.

PR #581, opened 2025-04-08
Fix docstring for XRootD DID finder

### LINKED ISSUE #570: XRootD dataset docstring is incorrect
Looks like it was a straight cut/paste from the open data guy.

```python
class XRootDDatasetIdentifier(DataSetIdentifier):
    def __init__(self, pattern: str, num_files: Optional[int] = None):
        r"""
        CERN Open Data Dataset - this will be looked up using the CERN Open Data DID finder.

        :param dataset: The dataset ID - this is an integer.
        :param num_files: Maximum number of files to return. This is useful during development
            to perform quick runs. ServiceX is careful to make sure it always
            returns the same subset of files.

        """
        super().__init__("xrootd", pattern, num_files=num_files)

    yaml_tag = "!XRootD"

    @classmethod
    def from_yaml(cls, _, node):
        return cls(node.value)
```

ponyisi:
Fix #570

### DIFF
--- a/servicex/dataset_identifier.py
+++ b/servicex/dataset_identifier.py
@@ -143,9 +143,9 @@ class CERNOpenDataDatasetIdentifier(DataSetIdentifier):
 class XRootDDatasetIdentifier(DataSetIdentifier):
     def __init__(self, pattern: str, num_files: Optional[int] = None):
         r"""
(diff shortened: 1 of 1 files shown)


gordonwatts (approved):
I hate when I discover these cut/paste codes. :-)

PR #588, opened 2025-05-08
Adding steps to pypi workflow - bot commit version bump

ArturU043:
I added a couple of steps in `pypi.yml` to commit the version bump when a release is created.

I also changed the version tag replacement using `sed`. 
There was an extra white space being introduced. This would cause the reggex search to fail on the next release event.

This should fix the doc building issue (currently showing the wrong release tag) automaticaly for next releases. 
This garantees `pyproject.toml` in master is consistent with the distributed package tag. 

I also manualy set the version to 3.1.1 since that release is published. 

--------------------------------------------------------------------------------

Tested with pre-releases on the feat/version-bump-gha branch, and the github bot correctly comits the version bump.

### DIFF
--- a/.github/workflows/pypi.yaml
+++ b/.github/workflows/pypi.yaml
@@ -43,8 +43,8 @@ jobs:
 
     - name: Update Version
       run: |
-        sed -i '/version =/ s/= "[^"][^"]*"/ = "${{ env.RELEASE_VERSION }}"/' pyproject.toml
-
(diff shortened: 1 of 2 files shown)


gordonwatts (commented):
These changes look good to me. I'd like Ben or Peter to also comment to make sure we all know what is going on in this logic. I don't think most of us are used to auto-updating packages during a release. I like it, because now when you check it out you get the right version in your code.

PR #591, opened 2025-05-14
Adding section to analysis utils package

ArturU043:
Added a section to the documentation about the parallel package with extra tools. 

The section does not go into details and points to the GH repo and the documentation.

I investigated the functionalities of the `intersphinx` extension. They could be useful if the second package grows and information has to be added more extensively to the client's doc. For now just pointing to links is enough. 

I placed this section after the CLI module.


BenGalewsky (approved):
Great, thanks

PR #592, opened 2025-05-15
add supported server resources checks

MattShirley:
Adds client support for checking whether a server supports a given resource/endpoint.

This PR shows two implementation patterns:
1. The actual ServiceXAdapter method that call a specific resource will throw an Exception/GuardList entry (instead of executing an HTTP request).
2. Using the resource name as a feature flag to code old/new conditional branches.

### DIFF
--- a/servicex/query_core.py
+++ b/servicex/query_core.py
@@ -560,23 +560,30 @@ class Query:
             if progress:
                 progress.advance(task_id=download_progress, task_type="Download")
 
+        transformation_results_enabled = "transformationresults" in await self.servicex.get_resources()
+
         while True:
             if not cached_record:
                 await asyncio.sleep(self.minio_polling_interval)
             if self.minio:
                 # if self.minio exists, self.current_status will too
                 if self.current_status.files_completed > len(files_seen):
-                    new_begin_at = datetime.datetime.now(tz=datetime.timezone.utc)
-                    files = await self.servicex.get_transformation_results(
-                        self.current_status.request_id, begin_at
-                    )
-                    begin_at = new_begin_at
+                    if transformation_results_enabled:
+                        new_begin_at = datetime.datetime.now(tz=datetime.timezone.utc)
+                        files = await self.servicex.get_transformation_results(
+                            self.current_status.request_id, begin_at
+                        )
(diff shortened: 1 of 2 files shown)


servicex/servicex_adapter.py
```diff
@@ -229,6 +351,7 @@ async def delete_transform(self, transform_id=None):
                         f"Failed to delete transform {transform_id} - {msg}"
                     )
 
+    @requires_resource("transformationresults")
```
ponyisi, line 354: In principle, in your code, we should never fail this check, right?

servicex/query_core.py
```diff
@@ -560,23 +560,32 @@ async def get_signed_url(
...
                 # if self.minio exists, self.current_status will too
                 if self.current_status.files_completed > len(files_seen):
-                    new_begin_at = datetime.datetime.now(tz=datetime.timezone.utc)
-                    files = await self.servicex.get_transformation_results(
-                        self.current_status.request_id, begin_at
-                    )
-                    begin_at = new_begin_at
+                    if transformation_results_enabled:
+                        new_begin_at = datetime.datetime.now(tz=datetime.timezone.utc)
+                        files = await self.servicex.get_transformation_results(
```
ponyisi, line 575: is there any reason that `get_transformation_results` can't just return the same list of objects that `self.minio.list_bucket()` would? This would simplify the failover logic

MattShirley: list_bucket returns a list of objects with a `filename` attribute, while the ServiceX adapter call is just turning JSON into dictionaries which have a 'file-path' key. I can change the key name to 'filename', but I don't think it's a good idea to change the dictionaries to objects with filename attributes given that pattern isn't reused anywhere else in the ServiceX adapter.

ponyisi: In principle you could of course change the minio adapter to list the files with a dictionary instead of a list of objects of course... I would just like the "list of files to be downloaded" to be a type that doesn't depend on the source of the information.

Note also my comment on the backend PR - I'm concerned that this thing may need a new column in the database anyway, and whether you want to keep the same name may be up for discussion.

PR #593, opened 2025-05-17
Add back concurrency semaphore

ponyisi:
I had misunderstood the concurrency limit in the boto3 configuration as a global limit, but it is only a limit on the number of parallel downloads for each individual download request. Therefore we could still be spawning a huge number of simultaneous S3 connections.

This PR puts semaphores back to limit the number of distinct simultaneous connections. It also tries to reduce the total number of connections made:
* it reuses metadata obtained in the bucket listing,
* it uses a lower level part of the library that means that fewer requests are made.

The new magic numbers are
* maximum 5 simultaneous bucket listings
* maximum 10 distinct files being downloaded simultaneously (configurable)
* maximum 5 download streams per file (internal boto configuration)

### DIFF
--- a/servicex/minio_adapter.py
+++ b/servicex/minio_adapter.py
@@ -38,15 +38,18 @@ import asyncio
 
 from servicex.models import ResultFile, TransformStatus
 
+# Maximum five simultaneous streams per individual file download
 _transferconfig = TransferConfig(max_concurrency=5)
-_sem = asyncio.Semaphore(10)
+# Maximum ten files simultaneously being downloaded (configurable with init_s3_config)
+_file_transfer_sem = asyncio.Semaphore(10)
+# Maximum five buckets being queried at once
 _bucket_list_sem = asyncio.Semaphore(5)
 
 
 def init_s3_config(concurrency: int = 10):
     "Update the number of concurrent connections"
-    global _sem
-    _sem = asyncio.Semaphore(concurrency)
+    global _file_transfer_sem
+    _file_transfer_sem = asyncio.Semaphore(concurrency)
 
 
 def _sanitize_filename(fname: str):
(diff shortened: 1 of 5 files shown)


ponyisi:
@gordonwatts I think this may be behind some of the stuttering you were seeing with the progress bars: a large number of simultaneous downloads could delay the status update requests for a very long time.

gordonwatts (commented):
I'm not an expert of the low-level boto workings. But it looks to me like:

* Up to 5 simultaneous bucket listings
* Up to 10 simultaneous downloads
* Up to 5 streams per download.

Is that right? Could you put the default config in the PR's text - just so that people going back have an idea.

I'm eager to see how this woks under various circumstances. I need to try this out again on a plane to see if it is robust against broken connections. :-)

servicex/minio_adapter.py
```diff
@@ -28,22 +28,25 @@
...
 
 import aioboto3
 from boto3.s3.transfer import TransferConfig
+import asyncio
 
 from servicex.models import ResultFile, TransformStatus
 
-_transferconfig = TransferConfig(max_concurrency=10)
+_transferconfig = TransferConfig(max_concurrency=5)
+_sem = asyncio.Semaphore(10)
```
BenGalewsky, line 42: These important magic numbers need  more comments to highlight their funciton. I also think `_sem` is not a great variable name

BenGalewsky (changes requested):
Welcome back, semaphores.. we missed you!
I think these important performance settings need to be highlighted with comments a bit better

servicex/minio_adapter.py
ponyisi: Renamed and comments added

gordonwatts (approved):
Thanks!

BenGalewsky:
How did all of this RNtuple code get in here.?

ponyisi:
@BenGalewsky branch merge I'm afraid.

BenGalewsky (approved):
Thanks for cleaning up the commits

PR #595, opened 2025-05-21
Revert "Bump astral-sh/setup-uv from 5 to 6 in the actions group"

ponyisi:
Reverts ssl-hep/ServiceX_frontend#584, I think this was a breaking change on all the tests


ponyisi:
Close in favor of #596

PR #597, opened 2025-05-22
Fail on authentication tokens without expiration

ponyisi:
After consulting some site experts, we do not expect anyone to deploy tokens with no expiration. Raise an error if such a token is encountered.


BenGalewsky (approved):
Yay, love this

PR #599, opened 2025-05-29
Changing databinder model Name limit

ArturU043:
Raising the 128-character limit on `Name` strings to 512. 

This would fix issues [15](https://github.com/ssl-hep/ServiceX_analysis_utils/issues/15) and [13](https://github.com/ssl-hep/ServiceX_analysis_utils/issues/15) from servicex_analysis_utils

Changed the databinder test to account for this new max length.

### DIFF
--- a/tests/test_databinder.py
+++ b/tests/test_databinder.py
@@ -16,7 +16,7 @@ def basic_spec(samples=None):
 
 
 def test_long_sample_name():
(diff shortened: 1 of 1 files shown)


ponyisi:
This would need to be coupled to a change in the backend database model - the limit in the frontend is only to ensure that the requests do not fail when actually submitted. Let's come back to this in a bit.

PR #601, opened 2025-06-06
Fix test issues on Windows

ponyisi:
There were some odd test failures on Windows. Identified a few issues with the mocking environments:
* needed to patch the token renewal path for an expired token
* S3 mocker actually tore down bucket state between tests and on Windows we would issue a delete command when the object was already gone


BenGalewsky (approved):
Thanks for cleaning up this annoyance!

PR #614, opened 2025-07-04
Determine sample name length limit from server capability

ponyisi:
Release 3.2.0 will not protect against submissions of too long sample names to pre-1.7.0 servers. Use a proper mechanism to determine whether the server should be able to handle long names.

### DIFF
--- a/servicex/servicex_adapter.py
+++ b/servicex/servicex_adapter.py
@@ -79,6 +79,7 @@ class ServiceXAdapter:
 
         # interact with _servicex_info via get_servicex_info
         self._servicex_info: Optional[ServiceXInfo] = None
+        self._sample_title_limit: Optional[int] = None
 
     async def _get_token(self):
         url = f"{self.url}/token/refresh"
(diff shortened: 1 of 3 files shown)


servicex/servicex_adapter.py
```diff
@@ -162,6 +163,22 @@ async def get_servicex_info(self) -> ServiceXInfo:
...
+    async def get_servicex_sample_title_limit(self) -> Optional[int]:
+        # check if the capability is defined
+        capabilities = await self.get_servicex_capabilities()
+        for capability in capabilities:
+            if capability.startswith("long_sample_titles_"):
+                try:
+                    # hope capability is of the form long_sample_titles_NNNN
+                    return int(capability[19:])
+                except ValueError:
+                    raise RuntimeError(
```
gordonwatts, line 175: If we have SX change how it delivers this, this will make it impossible to run against this SX. Perhaps, if there is an exception, then just move on, and return None if it isn't in the list of capabilities?

gordonwatts: Use `logging.warn` or similar?

servicex/servicex_client.py
```diff
@@ -183,7 +183,9 @@ def get_codegen(_sample: Sample, _general: General):
 
     sx = ServiceXClient(backend=servicex_name, config_path=config_path)
     datasets = []
+    title_length_limit = make_sync(sx.servicex.get_servicex_sample_title_limit)()
```
gordonwatts, line 186: Is this called inside a async? I'm worried that we move in and out of async here - the only reason this wasn't async was because we didn't use it that way. For example, a `async` version of `deliver`?

gordonwatts: Ok - this is called directly from `deliver`, and `deliver` for whatever reason has no async version. So... I guess this is needed.

gordonwatts (commented):
Looks good - there is one place where I think the code could be a little more defensive. And my standard issue with `deliver` not having a async version, which means everything has to be `make_sync`.

BenGalewsky (approved):
ok

PR #617, opened 2025-07-12
Fix codegen flush uses resolved backend

gordonwatts:
Update so that `servicex codegetn flush` uses the default backend name resolution used by the rest of `servicex`. This also prevents the silent behavior of command previously if you left off `-b <backend`.

## Summary
- maintain resolved backend name inside `ServiceXClient`
- use that backend for `servicex codegen flush`
- add regression test for default backend flush

## Notes

- The `backend` uses the internal client backend resolution - but if a `url` is specified instead, this might still be `None`. I'm not familiar enough with how the `url` is used to know what to do there.
- Anothe option is to require `-b` in the command line.
- And print an INFO message what backend's codegen was actually flushed.

## Testing
- `pytest -q`

------
https://chatgpt.com/codex/tasks/task_e_687268c2b26083209d70544ae44ddb70

### DIFF
--- a/servicex/app/codegen.py
+++ b/servicex/app/codegen.py
@@ -46,7 +46,9 @@ def flush(
     """
     sx = ServiceXClient(backend=backend, config_path=config_path)
     cache = sx.query_cache
-    cache.delete_codegen_by_backend(backend)
+    # Use the backend resolved by the ServiceXClient so that the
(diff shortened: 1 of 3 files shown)


ponyisi:
Largely superseded by #619 (gap is if we would want to keep the `sx.backend` property to track backends...)

PR #619, opened 2025-07-13
Remove codegen caching

ponyisi:
Codegen caching on the client side is not saving very much network activity but potentially causes users to be surprised that backend installations apparently don't support transformations that they should. In any case we obtain this data on every session because we request the capabilities, so we can actually eliminate the caching with zero added cost to the backend.


servicex/servicex_client.py
```diff
@@ -407,18 +407,10 @@ def cancel_transform(self, transform_id) -> None:
     def get_code_generators(self, backend=None):
         r"""
         Retrieve the code generators deployed with the serviceX instance
+        The backend argument is deprecated and has no effect
```
BenGalewsky, line 410: Do we think anyone outside of the cli has ever used this function? Seems not unreasonable to just delete this argument now

BenGalewsky (approved):
Just wonder about keeping that `backend` parameter in the servicex client method, but otherwise this is a better approach. Thanks

servicex/servicex_client.py
ponyisi: Yeah, that's cleaner. No need to be too cautious

PR #632, opened 2025-07-21
Extend timeout for httpx `put` in submission

ponyisi:
httpx has much shorter timeouts than aiohttp did, and large submissions can exceed the default 5s limits. Raise this for submissions to be much closer to aiohttpx.


BenGalewsky (approved):
Thanks

PR #659, opened 2025-09-24
Add dataset list name filtering

gordonwatts:
## Summary
- add an optional name filter argument to `servicex datasets list` that supports `*` wildcards and substring matching by default (use standard linux filename match rules from `fnmatch` python package)
- filter cached dataset results in the CLI before rendering the table
- cover the new behavior with CLI tests for both direct and wildcard patterns

------
https://chatgpt.com/codex/tasks/task_e_68cff0324a948320a0b6327245bb5e24


BenGalewsky (approved):
Thanks, looks useful

PR #662, opened 2025-09-27
Raise error when transform yields no files

gordonwatts:
## Summary
- raise a ServiceXException when a transform completes successfully but returns no files and provide a dataset identifier in the message
- extend dataset identifier classes to describe themselves so Query reuses those descriptions when building error messages, including file count previews for FileList datasets, and ensure every identifier subclass overrides the helper
- add coverage for the new behaviour and adjust existing dataset tests to avoid empty mocked downloads, including description coverage for CERN Open Data and XRootD identifiers

## Testing
- black servicex/query_core.py tests/test_dataset.py servicex/dataset_identifier.py tests/test_dataset_identifier.py
- flake8
- pytest

------
https://chatgpt.com/codex/tasks/task_e_68d5601c52648320850716e40a72cf75

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -45,7 +45,7 @@ jobs:
       - name: Install dependencies
         run: |
           uv venv
-          uv pip install --upgrade '.[test]'
+          uv pip install --upgrade '.[test,docs]'
           uv pip list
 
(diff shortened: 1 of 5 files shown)


gordonwatts:
Code will be in #657 .

PR #654, opened 2025-09-08
add display_results option to deliver and --hide-results flag to cli

MattShirley:
Adds a parameter to the `deliver` function that prints styled text via the `rich` lib. This is currently enabled by default for both direct Python code and the CLI (my decision, this PR can be adjusted to change that if needed).

To disable on Python:

```python
deliver(spec, display_results=False)
```

To disable with CLI:
```bash
servicex deliver examples/config_FuncADL_Uproot.yaml --hide-results
```

Example output:

```
% hatch run servicex deliver examples/config_FuncADL_Uproot.yaml
Delivering examples/config_FuncADL_Uproot.yaml to ServiceX cache
Uproot_FuncADL_YAML: Transform ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 3/3 00:35
                      Download ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 3/3 00:40

✓ ServiceX Delivery Complete!

                                                               Delivered Files
┏━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Sample              ┃ File Count ┃ Files                                                                                                  ┃
┡━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│ Uproot_FuncADL_YAML │          3 │ /private/var/folders/h6/k0_2vvjd3rz1tw9d0xw9t9sc0000gn/T/ServiceX_Client/cache-dir/679ef47c-2755-48f0… │
│                     │            │ /private/var/folders/h6/k0_2vvjd3rz1tw9d0xw9t9sc0000gn/T/ServiceX_Client/cache-dir/679ef47c-2755-48f0… │
│                     │            │ /private/var/folders/h6/k0_2vvjd3rz1tw9d0xw9t9sc0000gn/T/ServiceX_Client/cache-dir/679ef47c-2755-48f0… │
[...]

PR #675, opened 2025-10-16
Refactor to put rich output formatting in the app namespace

ponyisi:
#654 added result printout functionality directly into the `deliver()` function, which will change the behavior quite significantly for people using the API directly. Refactor this code to CLI responsibility.

### DIFF
--- a/tests/app/test_app.py
+++ b/tests/app/test_app.py
@@ -74,44 +74,6 @@ def test_deliver_hide_results(script_runner):
         assert result_rows[0] == "Delivering foo.yaml to ServiceX cache"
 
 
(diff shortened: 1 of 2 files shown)


servicex/app/main.py
```diff
@@ -63,6 +63,46 @@ def show_version(show: bool):
         raise typer.Exit()
 
 
+def _display_results(out_dict):
```
gordonwatts: Is this something others might find useful? e.g. should this get moved into the central library?

ponyisi: I was wondering that. For the moment I decided that `servicex.app` was that namespace...

PR #678, opened 2025-10-31
Fail the UprootRaw_Dict example if the server returns unexpected values

ponyisi:
Addresses https://github.com/ssl-hep/ServiceX/issues/1195


BenGalewsky (approved):
Great!

PR #680, opened 2025-11-04
support custom science images

MattShirley:
Adds `image` to selection criteria. Front end work for https://github.com/ssl-hep/ServiceX/issues/1162


ponyisi:
Perhaps just "image"?

ponyisi:
@MattShirley This may be something we need to fix - the client should definitely report an error. Should we not get a 500 status code?

ponyisi (commented):
See comment about `registry`

PR #689, opened 2025-11-26
Shift tag after doing version bump

ponyisi:
The release tags were created before the version bump was done, meaning that the tags still had the old release name in `pyproject.toml`. This meant the documentation had an off-by-one error in release numbers. Hopefully this fixes this.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "servicex"
-version = "3.2.3b"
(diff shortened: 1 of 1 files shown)


ArturU043:
I tested a dummy release from this branch and it works properly 
`pyproject.toml` is updated and matches the [release tag](https://github.com/ssl-hep/ServiceX_frontend/blob/refs/tags/3.2.3b/pyproject.toml) now.
Thanks!

PR #692, opened 2025-12-01
pin version of types-aiobotocore to <=2.26.0

MattShirley:
Pin version of `types-aiobotocore` to <=2.26.0 because the next version removes the `s3` subpackage.

This feature could be re-added with a new PyPI package (`types-aiobotocore-s3`), however compatibility with existing dependency `pytest-aioboto3` currently requires the `s3` subpackage: https://github.com/phillipuniverse/pytest-aioboto3/blob/main/pyproject.toml#L28


kyungeonchoi (approved):
Looks good to me. Thanks!

PR #697, opened 2025-12-30
WIP: Fix incomplete Minio downloads when downloading many files concurrently

kyungeonchoi:
### Story

- When downloading a large number of files (~3000) from MinIO/S3 backend, a small number of files (~10) are downloaded with an incorrect size mostly at the end. The size mismatch is small (less than 1% of file size) and can be either smaller and larger than the expected size.
- This happens only with download. Signed URLs works fine.
- Played with the options of the `MinioAdapter` class (e.g. `transferconfig` for aioboto3, concurrency setting (`_file_transfer_sem`) in the `download_file` function) but nothing fixes the issue.
- On the other hand the standard-alone download using CLI command - `servicex transforms download <request-id>` works fine.

### Updates

- Download files using a .part suffix and rename them to the final filename only after validating the file size.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "servicex"
-version = "3.3.0"
+version = "3.3.1b"
 description = "Python SDK and CLI Client for ServiceX"
 readme = "README.md"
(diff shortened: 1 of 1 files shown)


BenGalewsky:
Really a result of [ServiceX Issue 1318](https://github.com/ssl-hep/ServiceX/issues/1318)

PR #702, opened 2026-01-21
add basic check CLI command

### LINKED ISSUE #685: Add ServiceX Check CLI Command
Just verify that the cli has a valid token and can connect to server

MattShirley:
Checks all endpoints in `servicex.yaml` and verifies the connection with token.


servicex/app/check.py
```diff
@@ -0,0 +1,72 @@
...
+    if not config.api_endpoints:
+        console.print("[yellow]No ServiceX endpoints configured.[/yellow]")
+        return
+
+    console.print("\n[bold]Checking ServiceX endpoints...[/bold]\n")
+
+    results = {}
+
+    async def wrapped_verify_token(endpoint):
+        result = await verify_token(endpoint.endpoint, endpoint.token)
```
BenGalewsky: the Info endpoint in the ServiceX backend doesn't require authentication, so this is not a great test. We could add some authenticated call to the server, or for now just throw away the results from a benign endpoint that does check auth, like `/servicex/datasets`

BenGalewsky (changes requested):
These changes need unit tests

BenGalewsky (approved):
Looks good. Please fix the merge conflicts

Issue #710, opened 2026-02-07
Improve UX for working with local cache


BenGalewsky:
# Problem
I want to be able to easily access the output from prevous transforms without needing to go through the deliver function 

1. Find the transform result from the local cache database
2. Easily get back the file list

PR #713, opened 2026-02-09
Add code to reconstitute `deliver()` from a cache area

ponyisi:
Addresses #710 . New function `servicex.read_dir` (name is up for discussion) which gives a mock `deliver()` return value corresponding to data in a cache area. Assumes cache is in fact valid (i.e. doesn't validate that files or URLs actually work).


ArturU043:
Hi 

Is there a reason for wanting to set `local_preferred` to `Fasle`? 
If both URLs and local downloads are available, we could always default to local and clear up some of the code. 

Maybe I am missing something about this point.

PR #717, opened 2026-02-20
Update the kibana link code to use the new, cleaner Kibana URL

BenGalewsky:
Companion to [ServiceX PR 1295](https://github.com/ssl-hep/ServiceX/pull/1295) and part of fix for [Issue 1111](https://github.com/ssl-hep/ServiceX/issues/1111)

# Problem
The default Kibana URL we've been using in ServiceX deployments is more complicated than it needs to be and makes manipulating the filters a heroic task.

# Approach
We devised a cleaner URL which will now be vended by ServiceX and included in the status messages. This code matches the code in the backend for producing a link that shows just messages associated with a particular transaction on a particular ServiceX deployment.


ponyisi (approved):
OK

PR #734, opened 2026-04-23
add client version to custom HTTP header in servicex_adapter

MattShirley:
Add client version to HTTP header so the server can consume this information.

For issue: https://github.com/ssl-hep/ServiceX/issues/832

Backend work: https://github.com/ssl-hep/ServiceX/pull/1412


BenGalewsky (changes requested):
Please update tests to verify this

PR #735, opened 2026-05-05
Added ServiceX version number to docs.

RogerJanusiak:


RogerJanusiak:
Fixed broken links. Solves issue: #726.

docs/userguide/source/conf.py
```diff
@@ -17,7 +18,9 @@
     "Innovation in Software for High Energy Physics (IRIS-HEP)"
 )
 author = "Institute for Research and Innovation in Software for High Energy Physics (IRIS-HEP)"
-html_title = "ServiceX User Guide"
+release = _pkg_version("servicex")
```
MattShirley: Nitpick: Is it possible to reuse the release/version definitions between the reference/userguide sections?

RogerJanusiak: I will check!

MattShirley: > I will check!

The best location to define these is probably `docs/__init__.py` if you can get the imports to work.

RogerJanusiak:
@MattShirley Added a shared value between both docs. Might not be cleaner, but allows for easier editing of the format later.

MattShirley:
> @MattShirley Added a shared value between both docs. Might not be cleaner, but allows for easier editing of the format later.

Hmm... I don't know if this is necessarily better if you have to include the `sys.path` statement plus the `noqa` statement. I'm going to approve the PR and you can decide whether the old or new way is better long term.

MattShirley (approved):

PR #739, opened 2026-06-03
change `servicex init` command text

### LINKED ISSUE #1298: Login link that after login directs the user to the profile page
It would be nice that the user is able to directed to the profile page where they can copy their token. This would make it easier for new users as well as current users.

MattShirley:
Align `servicex init` text with changes in https://github.com/ssl-hep/ServiceX/pull/1469


BenGalewsky (approved):
Yep

